PR #4, opened 2017-02-08
One Sample ICA record parsed from ETS provided TRT-like XML.

agorina:
This inserts just one record. I am working on adding more.

### DIFF
--- a/reporting/data/sample-ica-grade5-ela-2016.sql
+++ b/reporting/data/sample-ica-grade5-ela-2016.sql
@@ -1,0 +1,148 @@
+/**
+ **  Sample data parsed from ETS provided TRT-like XML
+ */
+INSERT INTO asmt (id, natural_id, grade_id,type_id, subject_id, academic_year, name, label, version) VALUES
+  (1, 'SBAC)SBAC-ICA-FIXED-G5E-COMBINED-2017-Winter-2016-2017', 5, 1, 2, 2016, 'SBAC-ICA-FIXED-G5E-COMBINED-2017', 'Grade 5 ELA', '9831');
+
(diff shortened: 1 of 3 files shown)


reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -1,4 +1,3 @@
-
```
agorina: The real process should be to create a new script, but since nobody is using it yet I have decided to modify the existing one.

reporting/data/sample-ica-grade5-ela-2016.sql
```diff
@@ -0,0 +1,148 @@
...
+  (4, 1, 4, 2657.87221487352, 42.9833504368907, 3);
+
+INSERT INTO exam_item (exam_id, item_id, score, score_status, response) VALUES
+  (1, 60347, 1, 'SCORED', 'D'),
+  (1, 51719, 1, 'SCORED', 'B'),
+  (1, 59217, 1, 'SCORED', 'D'),
+  (1, 30901, 1, 'SCORED', 'C'),
+  (1, 30899, 1, 'SCORED', 'B'),
+  -- TODO: need to figure out how to handle this type of the responses. It is a question with two parts
+  (1, 30891, 1, 'SCORED', '&lt;itemResponse&gt;&lt;response id="EBSR1"&gt;&lt;value&gt;A&lt;/value&gt;&lt;/response&gt;&lt;response id="EBSR2"&gt;&lt;value&gt;D&lt;/value&gt;&lt;/response&gt;&lt;/itemResponse&gt;'),
```
agorina: I think this needs to be decoded before being saved. I will do it before I check this in. Here and everywhere below.

PR #5, opened 2017-02-08
Sample IAB and ICA

agorina:
This adds one student with a sample ICA and another student with an IAB.
There are two student groups, each with one student.
A user with the login 'dwtest@example.com' is assigned to both groups.

The data is parsed from the ETS TRT like samples.
I verified that item responses work with IRIS.
The scripts to load the data are not managed by Flyway and have to be loaded manually. 

I hope this is enough to get started with MySQL version of the DB.

### DIFF
--- a/reporting/data/sample-iab-grade4-math-oa-2016.sql
+++ b/reporting/data/sample-iab-grade4-math-oa-2016.sql
@@ -1,0 +1,77 @@
+/**
+ **  Sample data parsed from ETS provided TRT-like XML
+ */
+INSERT INTO asmt (id, natural_id, grade_id,type_id, subject_id, academic_year, name, label, version) VALUES
+  (2, '(SBAC)SBAC-IAB-FIXED-G4M-OA-MATH-4-Winter-2016-2017', 4, 2, 1, 2016, 'SBAC-IAB-FIXED-G4M-OA-MATH-4', 'MTH IAB G4 OperationsAlgebraicThinking', '9835');
+
+-- TODO: How to find a claim? Does it even exist?
+INSERT INTO claim (id, asmt_id, min_score, max_score, code) VALUES
+  (5, 2, null, null, 'OA');
+
+-- TODO: should min/max be taken from assmt packge?
+INSERT INTO item (id, claim_id, target_id, item_key, bank_key) VALUES
+  (2010,  5, null, 2010,  '200'),
+  (18943, 5, null, 18943, '200'),
+  (8906,  5, null, 8906,  '200'),
+  (2014,  5, null, 2014,  '200'),
+  (2024,  5, null, 2024,  '200'),
+  (13980, 5, null, 13980, '200'),
+  (29233, 5, null, 29233, '200'),
+  (2018,  5, null, 2018,  '200'),
+  (11443, 5, null, 11443, '200'),
+  (30075, 5, null, 30075, '200'),
+  (18804, 5, null, 18804, '200'),
+  (45230, 5, null, 45230, '200'),
+  (2002,  5, null, 2002, '200'),
+  (18461, 5, null, 18461, '200'),
+  (13468, 5, null, 13468, '200'),
+  (14461, 5, null, 14461, '200');
+
+INSERT IGNORE INTO district (id, name, natural_id) VALUES
+  (1, 'Sample District 1', '01247430000000');
+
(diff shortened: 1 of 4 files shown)


reporting/data/sample-ica-grade5-ela-2016.sql
```diff
@@ -0,0 +1,148 @@
...
+  (4, 1, 4, 2657.87221487352, 42.9833504368907, 3);
+
+INSERT INTO exam_item (exam_id, item_id, score, score_status, response) VALUES
+  (1, 60347, 1, 'SCORED', 'D'),
+  (1, 51719, 1, 'SCORED', 'B'),
+  (1, 59217, 1, 'SCORED', 'D'),
+  (1, 30901, 1, 'SCORED', 'C'),
+  (1, 30899, 1, 'SCORED', 'B'),
+  -- TODO: need to figure out how to handle this type of the responses. It is a question with two parts
+  (1, 30891, 1, 'SCORED', '&lt;itemResponse&gt;&lt;response id="EBSR1"&gt;&lt;value&gt;A&lt;/value&gt;&lt;/response&gt;&lt;response id="EBSR2"&gt;&lt;value&gt;D&lt;/value&gt;&lt;/response&gt;&lt;/itemResponse&gt;'),
```
agorina: Ignore this TODO, it was resolved.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -1,4 +1,3 @@
-
```
agorina: Normally I would create a new script with a new version, but since it has not been used I changed this one.

reporting/data/sample-ica-grade5-ela-2016.sql
```diff
@@ -65,13 +65,13 @@ INSERT INTO item (id, claim_id, target_id, item_key, bank_key) VALUES
   (54683, 4, null, 54683, '200');
 
 -- NOTE: ETS samples with anonymized data do not have names for school and district
-INSERT INTO district (id, name, natural_id) VALUES
+INSERT IGNORE INTO district (id, name, natural_id) VALUES
```
agorina: This allows for the data sample scripts to be executed in a diff order.

rcordeau: Do you anticipate getting duplicate keys or unique information if the scripts are executed in different orders?

agorina: I handcrafted the unique ids and tried to make sure that there is no conflict. This particular line will not insert if the value already exists, so it should work.

reporting/data/sample-iab-grade4-math-oa-2016.sql
```diff
@@ -0,0 +1,77 @@
...
+  (14461, 5, null, 14461, '200');
+
+INSERT IGNORE INTO district (id, name, natural_id) VALUES
+  (1, 'Sample District 1', '01247430000000');
+
+INSERT IGNORE INTO school (id, district_id, name, natural_id) VALUES
+  (1, 1, 'Sample School 1', '30664640124743');
+
+INSERT IGNORE INTO state (code) VALUES
+  ('SM');
```
malaffoon: What can i store in my brain to remember that 'SM' is the sample state? i.e. what does it stand for?

agorina: It is random :) Let me know what you like and I will change it.

One thing to note: while TRT spec includes State/District/School names I do not see them in the anonymized samples from ETS.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -112,121 +111,130 @@ CREATE TABLE IF NOT EXISTS item (
...
-  last_or_surname varchar(35) NOT NULL, 
-  first_name varchar(35) NOT NULL, 
-  middle_name varchar(35), 
-  gender_id tinyint NOT NULL, 
-  ethnicity_id tinyint NOT NULL, 
-  first_entry_into_us_school_at date, 
-  lep_entry_at date, 
-  lep_exit_at date, 
-  is_demo tinyint, 
+  id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
malaffoon: This diff view really sucks. I think we should push Jeff again about ... well, *any* other tool.

PR #7, opened 2017-02-15
Misc. changes based on the recent feedback from Smarter Balanced.

agorina:
Added a table to control application and schema relationship.

### DIFF
--- a/reporting/sql/V201702141487123176__misc_fixes.sql
+++ b/reporting/sql/V201702141487123176__misc_fixes.sql
@@ -1,0 +1,27 @@
+/**
+** Remove is_demo flag from the student table; the rdw is not supposed to receive demo records.
+** Remove exam location.
+** Make an exam's session not optional.
+**/
+USE reporting;
+
+CREATE TABLE application_schema_version (
+   major_version int UNIQUE NOT NULL
+);
+
+INSERT INTO application_schema_version (major_version) VALUES (0);
+
+ALTER TABLE student DROP COLUMN is_demo;
+
+ALTER TABLE iab_exam
+  DROP FOREIGN KEY fk__iab_exam__exam_location,
+  DROP COLUMN asmt_session_location_id,
+  MODIFY session_id varchar(128) NOT NULL;
+
+ALTER TABLE exam
+  DROP FOREIGN KEY fk__exam__exam_location,
+  DROP COLUMN asmt_session_location_id,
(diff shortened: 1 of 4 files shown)


reporting/sql/V201702141487123176__misc_fixes.sql
```diff
@@ -0,0 +1,27 @@
...
+** Remove is_demo flag from the student table; the rdw is not supposed to receive demo records.
+** Remove exam location.
+** Make an exam's session not optional.
+**/
+USE reporting;
+
+CREATE TABLE application_schema_version (
+   major_version int UNIQUE NOT NULL
+);
+
```
agorina: The idea is that the application will check this version to determine if it is compatible with the db schema.
@malaffoon, do you think we should have major and minor versions?

reporting/sql/V201702141487123176__misc_fixes.sql
```diff
@@ -0,0 +1,27 @@
...
+INSERT INTO application_schema_version (major_version) VALUES (0);
+
+ALTER TABLE student DROP COLUMN is_demo;
+
```
agorina: RDW is not supposed to receive/accept demo records.

reporting/sql/V201702141487123176__misc_fixes.sql
```diff
@@ -0,0 +1,27 @@
...
+ALTER TABLE iab_exam
+  DROP FOREIGN KEY fk__iab_exam__exam_location,
+  DROP COLUMN asmt_session_location_id,
+  MODIFY session_id varchar(128) NOT NULL;
+
```
agorina: We discussed that it is not delivered in TRT format and is not important for this phase.

reporting/sql/V201702141487123176__misc_fixes.sql
malaffoon: I'm waffling a bit on this: should we just have a string in there similar to software versioning, or do we break it up into major/minor? If we break it up i think we should have both major and minor: major controls backward compatibility while minor allows forward compatibility. OTOH, if we only ever query the minor version for informational purposes, why bother adding it?

agorina: You nicely repeated my chain of thoughts :) 
I guess I will keep just a major version of now.

PR #10, opened 2017-02-28
add import tables

malaffoon:


warehouse/sql/V201702271488215478__add_import.sql
```diff
@@ -0,0 +1,33 @@
+/**
```
malaffoon: i guess it doesn't matter that much but the naming convention for the files is redundant: the `%s` field includes the year/month/day.

PR #17, opened 2017-03-17
Added support for multiple ethnicity for a student.

agorina:


reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -153,14 +153,20 @@ CREATE TABLE IF NOT EXISTS student (
...
   lep_entry_at date,
   lep_exit_at date,
   is_demo tinyint,
   birthday date NOT NULL
  );
 
+CREATE TABLE IF NOT EXISTS student_ethnicity (
+  id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  ethnicity_id tinyint NOT NULL,
+  student_id bigint NOT NULL
```
malaffoon: no foreign key on this?

agorina: I thought we decided to not have FK in this case?
Do you think that I need both to Student and Ethnicity, or just a student?

PR #22, opened 2017-03-29
Added trait scores support.

agorina:


reporting/sql/V201702061486427095__initial_dml.sql
```diff
@@ -194,10 +194,6 @@ INSERT INTO math_practice (practice, description) VALUES
   (6, 'Attend to precision'),
   (7, 'Look for and make use of structure'),
   (8, 'Look for and express regularity in repeated reasoning');
-INSERT INTO item_trait_score (id, dimension) VALUES
```
malaffoon: Aren't you still going to need a table that has the normalized text associated with a trait score?

agorina: On the reporting site? Not sure. I can add it back. We also need to decide if this type of table needs to be handled by the migrator. 
I will add it back for now and we can address it later.

malaffoon: Oh, didn't notice this was reporting side. Yeah, guess we still have to figure that out.

PR #26, opened 2017-04-13
adding in max_points to the item table, and fixing data load scripts …

rcordeau:
…for the mart.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/26)


reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -133,6 +133,7 @@ CREATE TABLE IF NOT EXISTS item (
   allow_calc boolean,
   dok tinyint NOT NULL,
   difficulty float NOT NULL,
+  max_points int NOT NULL,
```
agorina: According to the SBAC spec it is "unsigned float". Could you please change the type to match the spec?

rcordeau: Done.

PR #31, opened 2017-04-18
Student groups - happy path.

agorina:


reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -204,15 +204,15 @@ CREATE TABLE IF NOT EXISTS student_ethnicity (
...
   id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
-  created_by varchar(255) NOT NULL,
-  school_id mediumint,
   name varchar(255) NOT NULL UNIQUE,
-  exam_from date,
-  exam_to date NOT NULL,
+  school_id mediumint NOT NULL,
+  school_year smallint NOT NULL,
   subject_id tinyint,
+  active tinyint NOT NULL,
```
jtreuting: If this is just true/false would a bit field make more sense?

agorina: Maybe, I need to research it more. The last time I researched BOOLEAN in MySQL it was the same as tinyint. But 1) I may be wrong about it; 2) it may be the case when we need to support null. If you know for sure that *bit* it better, I will changed it (here and in other places).

jtreuting: I'm not certain, but I think they are different in newer versions.  But in practice I think the space savings​ is only if there are multiple bit columns that can be stored in a single byte.

rcordeau: So far in MySQL 5.6 and 5.7, the boolean type is a synonym for tinyint

jtreuting: Boolean is a synonym for tinyint(1) but bit or bit(1) is its own data type since 5.0.3. 

Here are a couple references I found:
- http://stackoverflow.com/questions/25428842/bit1-or-tinyint-for-flags-in-mysql
- https://dev.mysql.com/doc/refman/5.7/en/storage-requirements.html#data-types-storage-reqs-numeric
- https://techhelplist.com/tech-tutorials/42-databases/65-mysql-performance-bit-vs-tinyint-vs-enum

reporting/data/sample-iab-grade4-math-oa-2016.sql
```diff
@@ -37,8 +37,8 @@ INSERT IGNORE INTO state (code) VALUES
 INSERT INTO student (id, ssid, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at, lep_entry_at, lep_exit_at, birthday) VALUES
   (2, '2222222222', 'LastName2', 'FirstName2', 'MiddleName2', 1, '2012-08-14', '2012-11-13', null, '2000-01-01');
 
-INSERT INTO student_group (id, created_by, school_id, name, exam_from, exam_to, subject_id) VALUES
-  (2, 'dwtest@example.com', 1, 'Test Student Group 2', null, '2017-08-01', 1);
+INSERT INTO student_group (id, created_by, created_at, school_id, school_year, name, subject_id, active) VALUES
```
malaffoon: created_at is a date, not a timestamp?

agorina: you are correct, fixing.

reporting/sql/V201702061486427077__initial_ddl.sql
agorina: I am still struggling with this :
- According to some comments in the above referenced links, as well as here https://www.safaribooksonline.com/library/view/high-performance-mysql/9781449332471/ch04.html, for Innodb engine the storage is the same.
- I see some negative comments about indexing bit, again I am suspecting it is not about bit(1), but still. 
Maybe we can chat after stand up?

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -204,15 +204,15 @@ CREATE TABLE IF NOT EXISTS student_ethnicity (
   student_id bigint NOT NULL
 );
 
-
 CREATE TABLE IF NOT EXISTS student_group (
```
agorina: Should the data mart even have 'active' flag? Or should we expect that the migrate will make sure that only active groups are in the data mart?

reporting/sql/V201702061486427077__initial_ddl.sql
jtreuting: Yeah, I'm not 100% confident it is a better approach.  Documentation seems to be a bit confusing.

PR #36, opened 2017-04-20
IAB aggregate report.

agorina:


reporting/data/prototype/student-group.sql
```diff
@@ -1,54 +1,53 @@
...
-#     THEN 1
-#   ELSE 0
-#   END AS has_exams
-
-# FROM user_student_group ug
-#   JOIN student_group g ON ug.student_group_id = g.id
-#   JOIN school sch ON sch.id = g.school_id
-# WHERE ug.user_login = 'user8';
-
+--   , CASE (SELECT 1
```
agorina: not related, just changed the comment pattern to be consistent.

reporting/data/prototype/student-group.sql
```diff
@@ -76,4 +75,31 @@ FROM student_group_membership gm
...
+  JOIN school sch ON sch.id = iab.school_id
+  JOIN asmt a ON a.id = iab.asmt_id
+  -- this is probably not needed
+  JOIN subject s ON s.id = a.subject_id
+WHERE
+  (sch.id IN (-1) OR sch.district_id IN (-1) OR 1 = 1)
+  AND iab.school_year = 2016
+  AND asmt_id in(2)
+  -- this is probably not needed
+  AND s.id = 1;
```
agorina: @jtreuting and @chdietz - does this look close to being right?

reporting/data/prototype/student-group.sql
```diff
@@ -76,4 +75,31 @@ FROM student_group_membership gm
...
+SELECT
+  min(completed_at) as date,
+  session_id,
+  count(*),
+  asmt_id,
+  avg(scale_score) avg_scale_score,
+  sum(is_category1) as is_category1,
+  sum(is_category2) as is_category2,
+  sum(is_category3) as is_category3,
+  sqrt(sum(scale_score_std_err*iab.scale_score_std_err)/count(*)) as avg_scale_score_std_err
```
agorina: @malaffoon - is this correct?

reporting/data/prototype/student-group.sql
```diff
@@ -76,4 +75,31 @@ FROM student_group_membership gm
...
+  sum(is_category3) as is_category3,
+  sqrt(sum(scale_score_std_err*iab.scale_score_std_err))/count(*) as avg_scale_score_std_err
+FROM iab_exam iab
+  JOIN student_group_membership gm ON iab.student_id = gm.student_id and gm.student_group_id = 55
+  JOIN school sch ON sch.id = iab.school_id
+  JOIN asmt a ON a.id = iab.asmt_id
+  -- this is probably not needed
+  JOIN subject s ON s.id = a.subject_id
+WHERE
+  (sch.id IN (-1) OR sch.district_id IN (-1) OR 1 = 1)
```
jtreuting: Didn't notice this on the last PR but i think it was the same there.

If I'm reading that right, even if school IDs are provided, it will always return all records since the `1 = 1` is always there.  Is this example just showing that use case, and in practice when school IDs or district IDs are provided the `1 = 1` will not be included?

agorina: In this case it means a user has a statewide access. 

This is discussed in this review: https://github.com/SmarterApp/RDW_Reporting/pull/59

I do not think we have settled on this approach (at least yet). But the idea is to have a 'template' SQL that will be populated at run time. The suggested template is : 
```
  (school.id IN (:schoolIds))  OR school.district_id IN (:districtIds)  OR 1 = :stateWide;
```

If this does not work, we will have to find a way to add this part to each SQL.

jtreuting: Right, for some reason I had in my mind that the `OR 1 = 1` would have to be dynamically added or kept off, but duh, the last bit can be templated.

reporting/data/prototype/student-group.sql
malaffoon: yes; although why does one have `iab.` prefix and one doesn't?

agorina: I obviously have a selecting attention and mix the prefix/no prefix thing. Will fix.
And yes, after the calculations the values are decimal.

PR #38, opened 2017-04-21
Change exam items to have a native item id.

agorina:
I am about to submit a corresponding ingest change.


reporting/data/sample-iab-grade4-math-oa-2016.sql
```diff
@@ -53,20 +53,20 @@ INSERT INTO iab_exam (id, school_year, asmt_id, asmt_version, opportunity, statu
 -- TODO: this needs more research.
```
agorina: Do people use these scripts? I would love to stop supporting it since we now can import data and 'migrate' them. Thoughts?

bgsavage: I would like to know as well if I need to maintain this script.

malaffoon: I personally i have never used them. But i don't actually run the product or anything.

PR #43, opened 2017-04-27
Mock up of the staging tables

agorina:
I am not planning on checking this in. Just wanted to show the flow before today's meeting.

It only shows the assessment and asmt items tables. It handles the current complexity of only pulling those that have exams. Does not handle deletes.

What I like about staging tables is that I can pull in duplicate data, then clean them up and 'bulk' update/insert.

Would like to know what people think.

### DIFF
--- a/reporting/data/prototype/migrate-hack.sql
+++ b/reporting/data/prototype/migrate-hack.sql
@@ -1,22 +1,245 @@
+USE reporting;
+
+-- TODO: staging tables should be in a different schema
+
+DROP TABLE IF EXISTS stage_import;
+CREATE TABLE IF NOT EXISTS stage_import (
+  id      BIGINT  NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  status  TINYINT NOT NULL,
+  content TINYINT NOT NULL,
+  message TEXT
+);
+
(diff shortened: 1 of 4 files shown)


chdietz:
you could check in a 'stage' prototype folder.

agorina:
@chdietz, after we talk, and if we agree on the approach - I can clean it up and check it in. Just do not want to do it prematurely.

reporting/data/prototype/migrate-hack.sql
```diff
@@ -1,22 +1,245 @@
...
+    a.id,
+    natural_id,
+    grade_id,
+    type_id,
+    subject_id,
+    a.school_year,
+    name,
+    label,
+    version
+  FROM warehouse.import i
```
chdietz: Are you thinking if we go with staging tables and pull method the 'process' that queries stage and inserts to the reporting mart will also need to query the warehouse?

agorina: Yes, we will need to get the data into staging table. 
For that we will need to query warehouse. 
Regardless of the 'pull' or 'push' approach, the code data is still in the warehouse.

PR #44, opened 2017-04-28
tweak groups to support working with them during exam processing

malaffoon:


reporting/data/prototype/migrate-hack.sql
```diff
@@ -94,8 +94,8 @@ CREATE PROCEDURE reporting.create_student_groups()
     REPEAT
       SELECT max(id) +1 INTO idVal FROM student_group;
       IF (idval is null) THEN SET idVal = 1; END IF;
-      INSERT INTO student_group (id, created_by, school_id, school_year, name, subject_id) VALUES
-        (idVal , 'dwtest@example.com', (SELECT id FROM school ORDER BY RAND() LIMIT 1), 2017, CONCAT('Test Student Group ', idVal), null);
+      INSERT INTO student_group (id, school_id, school_year, name, subject_id) VALUES
```
malaffoon: since created_by can be null, no point in faking it out

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -211,22 +211,24 @@ CREATE TABLE IF NOT EXISTS student_group (
   school_id mediumint NOT NULL,
   school_year smallint NOT NULL,
   subject_id tinyint,
-  created_by varchar(255) NOT NULL,
-  created_at timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  creator varchar(250),
```
malaffoon: make consistent with import table
make creator optional

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -211,22 +211,24 @@ CREATE TABLE IF NOT EXISTS student_group (
...
+  created timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
   CONSTRAINT fk__student_group__school FOREIGN KEY (school_id) REFERENCES school(id),
   CONSTRAINT fk__student_group__subject FOREIGN KEY (subject_id) REFERENCES subject(id)
 );
 
 CREATE TABLE IF NOT EXISTS student_group_membership (
   student_group_id int NOT NULL,
   student_id bigint NOT NULL,
+  CONSTRAINT uk__student_group_id__student_id UNIQUE KEY (student_group_id, student_id),
```
malaffoon: i'm using this unique constraint to avoid duplicate entries

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -234,27 +234,29 @@ CREATE TABLE IF NOT EXISTS student_ethnicity (
 -- TODO: consider having a log of changes to the group: who did the change, when and what was changed
 CREATE TABLE IF NOT EXISTS student_group (
   id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
-  name varchar(255) NOT NULL UNIQUE,
```
malaffoon: i don't think a group's name has to be unique across all schools and years
i didn't add a compound key for uniqueness but i did think about it, e.g. UNIQUE KEY (name, school_id, school_year)

agorina: Why did you remove it?

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -234,27 +234,29 @@ CREATE TABLE IF NOT EXISTS student_ethnicity (
...
+  name varchar(255) NOT NULL,
   school_id mediumint NOT NULL,
   school_year smallint NOT NULL,
   subject_id tinyint,
   active tinyint NOT NULL,
-  created_by varchar(255) NOT NULL,
-  created_at timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  creator varchar(250),
+  created timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
   CONSTRAINT fk__student_group__school FOREIGN KEY (school_id) REFERENCES school(id),
```
malaffoon: hmm, wonder if i should add an index on `name`

warehouse/sql/V201702061486427077__initial_ddl.sql
malaffoon: I think the odds that two schools will want to use the same group names are pretty high. You know they're all going to have "Grade 3 Math" or "English 1a".

agorina: I agree, but we do need to add some uniqueness in there. As you are suggesting above. We do not want a teacher to have X groups with the same name.

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: If you add one for (name, school_id, school_year) that should work for now.
I am not sure how one is going to search for Groups.

warehouse/sql/V201702061486427077__initial_ddl.sql
chdietz: The name of the group should not be the unique identifier.  It is user defined.

chdietz: agree

PR #45, opened 2017-04-29
Feature/migrating to use gradle

rcordeau:
changing around how we execute migrations. Moving to use gradle provides, hopefully, a consistent experience across projects, and integrates well with projects that incorporate this project as a subproject.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/45)


reporting/build.gradle
```diff
@@ -0,0 +1,8 @@
+flyway {
+    url = 'jdbc:mysql://localhost:3306/'
+    user = 'root'
+    password = ''
+    schemas = ['reporting']
+    locations = ['filesystem:' + rootDir.path +'/reporting/sql']
```
rcordeau: This is done this way, with the rootDir.path added, to provide an absolute path to the SQL files for migration. Otherwise, if this project is used as part of a parent project (like RDW_Ingest), the parent project will execute the flywayMigrate (or clean) with an incorrect location relative to the parent directory, and not the actual subdirectory.

build.gradle
```diff
@@ -0,0 +1,28 @@
...
+apply plugin: 'org.flywaydb.flyway'
+
+subprojects {
+    apply plugin: 'org.flywaydb.flyway'
+
+}
+
+//'export' the subproject tasks so that they are easier to invoke, and can be invoked from
+//parent project
+task('migrateWarehouse', dependsOn: ':warehouse:flywayMigrate')
```
malaffoon: Why this approach as opposed to just using the tasknames as is? We all should know enough gradle to be able to specify targets. I suppose we could rename them from `flywayMigrate` to just `migrate`. 

Curious, if i do `gradle flywayClean` from the root will it do it for both warehouse and reporting? If so is that really worse than `gradle cleanAll`?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Fri Apr 28 17:14:58 PDT 2017
+distributionBase=GRADLE_USER_HOME
```
malaffoon: What are these for? I don't have a gradle user home explicitly defined anywhere.

reporting/flyway.properties
```diff
@@ -1,133 +0,0 @@
-# Jdbc url to use to connect to the database
```
malaffoon: Is there no way to keep this around and use it instead of specifying details in the build.gradle. Currently, this file can be used when executing flyway directly, in case you're trying to do something more complex than just clean/migrate (e.g. building a baseline, playing with repair/validate).

rcordeau: Sadly, no. The Gradle plugin won't read the flyway.properties file. It will read either the build.gradle file for properties, or a gradle.properties file. Also, the gradle plugin can do the same stuff as the command line, I just didn't 'export' those commands up:
From: https://flywaydb.org/documentation/gradle/
flywayMigrate  Migrates the database
flywayClean	Drops all objects in the configured schemas
flywayInfo	Prints the details and status information about all the migrations
flywayValidate	Validates the applied migrations against the ones available on the classpath
flywayBaseline	Baselines an existing database, excluding all migrations up to and including baselineVersion
flywayRepair	Repairs the metadata table

We can use the build.gradle file in the subdirectories (reporting, warehouse) as documentation just like that file for what settings are honored.  Any settings that are specified in the file can be overridden on the command line. System properties override Gradle properties override Plugin configuration.

gradle/wrapper/gradle-wrapper.properties
rcordeau: The distribution base specifies whether the unpacked wrapper distribution should be stored in the project or in the gradle user home dir. I upgraded from Gradle 3.3 to 3.5 and I think this was an automatic add from that process.

build.gradle
rcordeau: I had to use it this way because Warehouse and Reporting will likely have different schemas in the future, so they both need their own Flyway configuration (which schema and which .sql file migration locations). A lot of this stemmed from me wanting to be able to call a migrate or clean (or whatever) from RDW_Ingest, which has RDW_Schema as a subproject. From the RDW_Schema project, you could just do "./gradlew :warehouse:flywayMigrate", but this wouldn't work from the RDW_Ingest project, because it would look for :warehouse as a direct subdirectory/subproject of RDW_Ingest. Calling "./gradlew flywayClean or flywayMigrate) from he root will no do both warehouse and reporting - if it was configured, it would want to run flyway<command> on the one configuration, hence the clean all.  Also, configuration the flyway properties in the top level build.gradle file wasn't working as expected - if you define the global configuration and have tasks try to override those configurations, what Gradle actually ends up doing (surprisingly) is pre-processing all of the task definitions and applying all of the overrides, ending up with the last override, which "wins" and becomes the global setting for every task, regardless of what overrides are supposed to be applied. 
Since there is no "RDW_Schema" top-level flyway configuration, calling ./gradlew flywayClean won't do anything - the configurations are all in the subdirectories. The Gradle plugin used to have support for multiple configurations a while ago, but that got removed. The recommendation is separate flyway configurations and separate flyway executions.

README.md
```diff
@@ -8,27 +8,31 @@ brew install flyway
...
 OR
-warehouse$ ../scripts/clean
+RDW_Schema$ ./gradlew cleanAll
 ```
 
 #### Alternate Data Source
 The data source, user, password, etc. can be overridden on the command line, e.g.
 ```bash
-warehouse$ ../scripts/migrate -url="jdbc:mysql://rdw-aurora-dev.cugsexobhx8t.us-west-2.rds.amazonaws.com:3306/" -user=sbac -password=mypassword
+RDW_Warehouse$ ./gradlew -Pflyway.url="jdbc:mysql://rdw-aurora-dev.cugsexobhx8t.us-west-2.rds.amazonaws.com:3306/" -Pflyway.user=sbac -Pflyway.password=mypassword
```
malaffoon: there should probably be an actual gradle task in the example here?

rcordeau: OK

build.gradle
malaffoon: ok

gradle/wrapper/gradle-wrapper.properties
malaffoon: ok

README.md
```diff
@@ -8,27 +8,31 @@ brew install flyway
 ```
 
 #### To create the schema 
-There are multiple schemas: a data warehouse and data mart(s). Each has a corresponding folder.
```
malaffoon: We talked offline about perhaps adding examples for invoking the other flyway operations. Don't need a lot of fluff documentation, just an example invocation for 'repair', 'validate', 'baseline', etc.

reporting/build.gradle
```diff
@@ -0,0 +1,8 @@
+flyway {
```
malaffoon: couldn't this be in the root build.gradle in a `subprojects`? just thinking about reducing duplication.

reporting/flyway.properties
malaffoon: ok

reporting/build.gradle
rcordeau: Yeah, not all of it, but the common parts can be.

PR #46, opened 2017-05-01
Added import_id to support migrate.

agorina:


warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -297,6 +303,7 @@ CREATE TABLE IF NOT EXISTS iab_exam (
   scale_score float,
   scale_score_std_err float,
   completed_at timestamp(0) NOT NULL,
+  import_id bigint NOT NULL,
```
agorina: Not a real change here, just my OCD made me move it.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -362,6 +368,7 @@ CREATE TABLE IF NOT EXISTS exam (
   scale_score_std_err float,
   achievement_level tinyint,
   completed_at timestamp(0) NOT NULL,
+  import_id bigint NOT NULL,
```
agorina: Same as above.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -429,19 +440,36 @@ CREATE PROCEDURE student_upsert (IN  p_ssid                          VARCHAR(65)
...
+      -- check if there is anything to update
+      SELECT CASE WHEN count(*) > 0 THEN 0 ELSE 1 END INTO isUpdate FROM student
+      WHERE id = p_id
+            AND last_or_surname = p_last_or_surname
+            AND first_name = p_first_name
+            AND middle_name = p_middle_name
+            AND gender_id = p_gender_id
+            AND first_entry_into_us_school_at <=> p_first_entry_into_us_school_at
+            AND lep_entry_at <=> p_lep_entry_at
+            AND lep_exit_at <=> p_lep_exit_at
```
agorina: The above three dates are Nullable, hence a null safe comparison.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -504,16 +536,29 @@ CREATE PROCEDURE school_upsert(IN  p_district_name       VARCHAR(100),
...
+          district_id = p_district_id,
+          import_id   = p_import_id
+        WHERE id = p_id;
+
+      END IF;
+
     ELSE
-      INSERT INTO school (district_id, name, natural_id)
-      VALUES (p_district_id, p_name, p_natural_id);
+      INSERT INTO school (district_id, NAME, natural_id, import_id)
```
agorina: I have no idea why 'name' has changed.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -220,8 +224,9 @@ CREATE TABLE IF NOT EXISTS student (
   first_entry_into_us_school_at date,
   lep_entry_at date,
   lep_exit_at date,
-  is_demo tinyint,
-  birthday date NOT NULL
+  birthday date NOT NULL,
```
jtreuting: Did you mean to drop `is_demo`?  Assuming yes, but checking.

warehouse/sql/V201702061486427077__initial_ddl.sql
jtreuting: wow, I never knew that existed.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -429,19 +440,36 @@ CREATE PROCEDURE student_upsert (IN  p_ssid                          VARCHAR(65)
...
+            AND last_or_surname = p_last_or_surname
+            AND first_name = p_first_name
+            AND middle_name = p_middle_name
+            AND gender_id = p_gender_id
+            AND first_entry_into_us_school_at <=> p_first_entry_into_us_school_at
+            AND lep_entry_at <=> p_lep_entry_at
+            AND lep_exit_at <=> p_lep_exit_at
+            AND birthday = p_birthday;
+
+      IF (isUpdate = 1)
```
jtreuting: Not sure the performance implications, but I know there is a `INSERT INTO ... ON DUPLICATE KEY UPDATE` command that can combine this into a single query instead of looking up the count first in order to do a switch on it.

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: Yes, thank you for asking. I meant to comment on it and forgot. 
According to Brandt we should not be receiving it and if we do - it is okay to reject. There is an outstanding JIRA (or a comment in one) to implement it.  Since it was not used anywhere I decided to remove.

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: Yes, you are right. But I have learned (the hard way) that with the InnoDB engine, if there is an auto-increment id, it removes it and creates a new one. 

https://dev.mysql.com/doc/refman/5.6/en/insert-on-duplicate.html
```(The effects are not identical for an InnoDB table where a is an auto-increment column. With an auto-increment column, an INSERT statement increases the auto-increment value but UPDATE does not.)```

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: I learned it today too :)

PR #47, opened 2017-05-01
Added import data to support migrate.

agorina:
I am not sure about this. I think it will be refactored once we start working on the real migrate. At this point it is a guess.

### DIFF
--- a/reporting/data/prototype/migrate-hack.sql
+++ b/reporting/data/prototype/migrate-hack.sql
@@ -1,22 +1,22 @@
 INSERT INTO reporting.district(id, name, natural_id)
   SELECT id, name, natural_id FROM warehouse.district;
 
-INSERT INTO reporting.school(id, district_id, name, natural_id)
-  SELECT id, district_id, name, natural_id FROM warehouse.school;
+INSERT INTO reporting.school(id, district_id, name, natural_id, import_id)
+  SELECT id, district_id, name, natural_id, import_id FROM warehouse.school;
 
-INSERT INTO reporting.student (id, ssid, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at, lep_entry_at, lep_exit_at, birthday)
-  SELECT id, ssid, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at, lep_entry_at, lep_exit_at, birthday FROM warehouse.student;
+INSERT INTO reporting.student (id, ssid, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at, lep_entry_at, lep_exit_at, birthday, import_id)
+  SELECT id, ssid, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at, lep_entry_at, lep_exit_at, birthday, import_id FROM warehouse.student;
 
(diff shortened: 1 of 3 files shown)


reporting/data/prototype/migrate-hack.sql
```diff
@@ -75,40 +77,15 @@ INSERT INTO reporting.exam (id, school_year,  asmt_id, asmt_version, opportunity
         AND m.num = 4
       ) AS claim4 ON claim4.exam_id = e.id;
 
-
 INSERT INTO reporting.exam_item (id, exam_id, item_id, score, score_status, response, position)
   SELECT id, exam_id, item_id, round(score), score_status, response, position FROM warehouse.exam_item;
 
--- TODO:  need an intelligent way to combine student into groups; something like given me students in the same grade that have the same session ids..?
```
agorina: Since this TODO is done there is not need to fake the groups.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -15,6 +15,22 @@ CREATE TABLE application_schema_version (
...
+
+CREATE TABLE IF NOT EXISTS import_migrate (
+  id bigint NOT NULL PRIMARY KEY,
+  status tinyint NOT NULL,
+  first_import_id bigint NOT NULL,
+  last_import_id bigint NOT NULL,
+  created timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  updated timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  message text
+ );
```
agorina: Would like to see some feedback if this makes sense.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -71,6 +87,7 @@ CREATE TABLE IF NOT EXISTS asmt (
   name varchar(250),
   label varchar(255),
   version varchar(30),
+  import_id bigint NOT NULL,
```
agorina: I added import_id to capture it but did not create a foreign key:
1.  (at least for now) there is not import table in the reporting DB
2.  i think for the auditing purpose it is good to keep track of this id here, but we will not be using it anywhere.

reporting/sql/V201702061486427095__initial_dml.sql
```diff
@@ -7,6 +7,10 @@ USE reporting;
 
 INSERT INTO application_schema_version (major_version) VALUES (0);
 
+INSERT INTO import_migrate_status (id, name) VALUES
+  (-100, 'REQUESTED'),
+  (100, 'MIGRATED');
```
agorina: Better names?

malaffoon: Could look to spring batch status as inspiration, e.g. STARTED, FAILED, ABANDONED, COMPLETED

agorina: I like it.

reporting/sql/V201702061486427077__initial_ddl.sql
malaffoon: i think having the `import_` prefix is unnecessary; migration is migration, we just happen to use import-id's to keep track of things. Also, do we need to normalize status?

Of course, your comment was probably referring to substantive feedback not minor nits. And i've got nothing until we sit down and start coding it.

PR #48, opened 2017-05-02
Schema int type and NULL clean up .

agorina:
This came up during one of the PR and I have reviewed/updated ```int``` data types and NULL fields.

For now I am only changing ```warehouse``` to solicit the feedback. I will align ```reporting``` once we are good with it.

Please provide feedback.

To save you a look up: 
ints


warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -80,21 +80,23 @@ CREATE TABLE IF NOT EXISTS gender (
 );
 
 CREATE TABLE IF NOT EXISTS accommodation (
-  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
-  code varchar(25) NOT NULL UNIQUE
+  id smallint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  code varchar(25) NOT NULL UNIQUE,
+  import_id bigint NOT NULL,
```
agorina: Unrelated, I missed it last time.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -80,21 +80,23 @@ CREATE TABLE IF NOT EXISTS gender (
...
-  id  bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  id smallint NOT NULL AUTO_INCREMENT PRIMARY KEY,
   natural_id varchar(250) NOT NULL UNIQUE,
   grade_id tinyint NOT NULL,
   type_id tinyint NOT NULL,
   subject_id tinyint NOT NULL,
   school_year smallint NOT NULL,
   name varchar(250),
-  label varchar(255),
+  label varchar(255), -- TODO: change this to NOT NULL when we do not need a hack code that creates assmt
```
agorina: I know we do not like ```TODO```, but I find it the best way to remember to do it. Let me know if we want to approach this differently.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -114,7 +116,7 @@ CREATE TABLE IF NOT EXISTS asmt_score (
 );
 
 CREATE TABLE IF NOT EXISTS claim (
-  id smallint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  id mediumint NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
agorina: Is this right?

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -123,7 +125,7 @@ CREATE TABLE IF NOT EXISTS claim (
 );
 
 CREATE TABLE IF NOT EXISTS subject_claim_score (
-  id smallint NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
agorina: I removed AUTO_INCREMENT for everything we have 'hardcoded' to upload ahead of time.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -313,9 +315,9 @@ CREATE TABLE IF NOT EXISTS iab_exam_item (
   id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
   iab_exam_id bigint NOT NULL,
   item_id int NOT NULL,
-  score float,
+  score float NOT NULL,
```
agorina: My understanding is that this cannot be null?

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -313,9 +315,9 @@ CREATE TABLE IF NOT EXISTS iab_exam_item (
...
   score_status varchar(50),
-  position int,
+  position int NOT NULL,
```
agorina: Ditto?

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -328,17 +330,16 @@ CREATE TABLE IF NOT EXISTS iab_exam_item (
 
 CREATE TABLE IF NOT EXISTS iab_exam_available_accommodation (
   iab_exam_id bigint NOT NULL,
-  accommodation_id int NOT NULL,
-  CONSTRAINT fk__iab_exam_available_accommodation__iab_exam FOREIGN KEY (iab_exam_id) REFERENCES iab_exam(id),
-  CONSTRAINT fk__iab_exam_available_accommodation__accomodation FOREIGN KEY (accommodation_id) REFERENCES accommodation(id)
```
agorina: Since we have decided to not enforce this type of constraints here, I am removing it.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -328,17 +330,16 @@ CREATE TABLE IF NOT EXISTS iab_exam_item (
...
+  accommodation_id smallint NOT NULL,
+  CONSTRAINT fk__iab_exam_available_accommodation__iab_exam FOREIGN KEY (iab_exam_id) REFERENCES iab_exam(id)
 );
 
 /** ICA and Summative exams **/
 
 CREATE TABLE IF NOT EXISTS exam_student (
   id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
agorina: Should this be ```int``? Since this represents ICA/Summative only?

warehouse/sql/V201702061486427077__initial_ddl.sql
malaffoon: I have no objection to TODO's in the code ... as long as we have a task to remove them all before release.

agorina: I will add either a new task or a comment to an existing one.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -80,21 +80,23 @@ CREATE TABLE IF NOT EXISTS gender (
...
+  code varchar(25) NOT NULL UNIQUE,
+  import_id bigint NOT NULL,
+  CONSTRAINT fk__accommodation__import FOREIGN KEY (import_id) REFERENCES import(id)
 );
 
 /** Assessment Packages related data **/
 
 CREATE TABLE IF NOT EXISTS asmt (
-  id  bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  id smallint NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
malaffoon: I can't say i agree with downsizing ids. You might convince me that `int` is okay here but asserting the we'll never end up with more than 65k ids when considering inserts, deletes, updates, reloading data, etc. just seems risky. And to what end? Saving a few thousands bytes of disk space?

warehouse/sql/V201702061486427077__initial_ddl.sql
malaffoon: I think we should not use `mediumint` at all. This is an auto-incremented id on a table with limited records, so use `int`.

malaffoon:
I guess i missed the initial conversation. I think this is not a good choice. We should use `bigint` or `int` for auto-incremented id fields. We can use smaller integer types for well-known values (like you did with removing auto-increment on lookup tables).

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: From a _smart_ book: "In general, try to use the smallest data type that can correctly store and represent your data. Smaller data types are usually faster, because they use less space on the disk, in memory, and in the CPU cache. They also generally require fewer CPU cycles to process. Make sure you don’t underestimate the range of values you need to store, though, because increasing the data type range in multiple places in your schema can be a painful and time-consuming operation. If you’re in doubt as to which is the best data type to use, choose the smallest one that you don’t think you’ll exceed."

I do not think the above conflicts with your comment. The question is what we consider "the smallest one that we will not exceed" .  The thought behind the change was: 12 asmts per subject per year X 5 subjects X 5 years = 300. Even with multiple attempts to update/delete, will we ever go above 65k? 

Keep in mind that we also JOIN on this field, so having a small value here is potentially beneficial. 

Having said all of the above, I think we should chat :)

warehouse/sql/V201702061486427077__initial_ddl.sql
jtreuting: I've always been under the impression that you use the smallest that you can because it gives the DB flexibility in storage, memory, etc and it can help performance (I don't have hard data to back that up though).

I think with `mediumint` it would be an `Integer` on the java side so I don't think it makes the code "weird" or anything.  So I don't personally have an issue with `mediumint`.

warehouse/sql/V201702061486427077__initial_ddl.sql
jtreuting: You are right.  If it's not score it's -1

warehouse/sql/V201702061486427077__initial_ddl.sql
jtreuting: position should not be null either

warehouse/sql/V201702061486427077__initial_ddl.sql
jtreuting: I would think so.  2 types x 5 subjects x 7 mil x 5 years = 350 mill.   With churn year after year that may double and int is 2 billion.  Also, realistically there are 2 subjects with a potential 3rd on the horizon which would be years away.

warehouse/sql/V201702061486427077__initial_ddl.sql
jtreuting: i do think downsizing is a good thing to do and agree with Alla that her comments don't contradict what Mark said either.  I'm not sure I can see getting close to 65k though.

In Alla's calculation there is also grade to consider so probably 15 assessments (ICA and Summative also) x 5 subjects x 5 years x 7 grades = 2,625

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: @jtreuting and @malaffoon, I think Mark's concern is discussed here:

https://groups.google.com/forum/?nomobile=true#!topic/comp.databases.mysql/b83SotGM7BQ

and at bottom of this chain:
https://github.com/github/gh-ost/issues/353

In short, there are conflicting data whether 3 byte int is converted to 4 byte int. 
Since there is no strong evidence that ```mediumint``` make sense I am inclining to use ```int``` instead of it.

I am going to abandon this review and create a new one that addresses all the discussed changes.

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: @malaffoon, after giving it more thought, I think I will change this to ```int```, unless you strongly disagree.

warehouse/sql/V201702061486427077__initial_ddl.sql
chdietz: So the db engine may pad the 3 bytes to 4?

warehouse/sql/V201702061486427077__initial_ddl.sql
malaffoon: If we document the limitation.

warehouse/sql/V201702061486427077__initial_ddl.sql
rcordeau: I know we aren't using MariaDB, but it might be close to what we will be using. MariaDB specifically calls out that MEDIUM ints will require 4 bytes of memory:
"Note that MEDIUMINT columns will require 4 bytes in memory (for example, in InnoDB buffer pool)." https://mariadb.com/kb/en/mariadb/data-type-storage-requirements/

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: Yes, will do. I am assuming that documenting in the code is enough. Please let me know if you think we should capture it somewhere else.

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: @chdietz , yes that is the concern. According to some various blogs it is not obvious. Some say that it does it under some conditions (for example during storage it uses 3 bytes but during computations converts to 4). But I could not find a strong argument that there are clear benefits to using it.

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: After I convinced @malaffoon that it should be ```smallint``` and then considered Mark's arguments, I have decided to go with just ```int``` here :) 

Once again, please speak up if you disagree.

PR #49, opened 2017-05-03
Schema int type and NULL clean up .

agorina:
This is the final change based on https://github.com/SmarterApp/RDW_Schema/pull/48

Main changes:
- asmt and student ids from BIGINT to INT
- added NOT NULL or comments to change it later
- replaced all MEDIUMINT with INT

After going back and forth on the changes to the id in the exam I kept it as is.
What made me decide this way: 
1. multiple discussions with diff.people
2. while I was putting comments explaining why exam id is of a different type than iab exam id, this just did not feel right.
3. After I looked at the data mart I remembered that the exam tables are highly de-normalized. So joining by this ID should be very limited - only to get an exam items.  
4. The fact that in Java one will be Long and another Integer.
My 'gut feeling' told me that maintenance of this outweighs a _potential_ benefit of performance.

As always, open to comments.


reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -71,14 +72,15 @@ CREATE TABLE IF NOT EXISTS gender (
 );
 
 CREATE TABLE IF NOT EXISTS accommodation (
-  id int NOT NULL PRIMARY KEY,
-  code varchar(25) NOT NULL UNIQUE
+  id smallint NOT NULL PRIMARY KEY,
+  code varchar(25) NOT NULL UNIQUE,
+  import_id bigint NOT NULL
```
agorina: Unrelated, forgot to add it in the prev. PR.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -182,16 +184,17 @@ CREATE TABLE IF NOT EXISTS item_difficulty_cuts (
...
 );
 
 CREATE TABLE IF NOT EXISTS school (
-  id mediumint NOT NULL PRIMARY KEY,
-  district_id mediumint NOT NULL,
-  name varchar(100) NOT NULL,
+  id int NOT NULL PRIMARY KEY,
+  district_id int NOT NULL,
   natural_id varchar(40) NOT NULL UNIQUE,
+  name varchar(100) NOT NULL,
```
agorina: Just moved around to match the warehouse schema. 
Makes it easier to compare two files.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -212,22 +216,21 @@ CREATE TABLE IF NOT EXISTS student (
   first_entry_into_us_school_at date,
   lep_entry_at date,
   lep_exit_at date,
-  is_demo tinyint,
   birthday date NOT NULL,
```
agorina: This was removed from the warehouse before.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -215,7 +217,7 @@ CREATE TABLE IF NOT EXISTS state (
 /** Student Groups */
 
 CREATE TABLE IF NOT EXISTS student (
-  id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
agorina: This requires a change to the Exam Processor. I will have a pull request shortly.

warehouse/sql/V201702061486427095__initial_dml.sql
```diff
@@ -77,75 +77,6 @@ INSERT INTO gender (id, name) VALUES
   (1, 'Male'),
   (2, 'Female');
 
-INSERT INTO accommodation (code) VALUES
```
agorina: This is not used.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -71,14 +72,15 @@ CREATE TABLE IF NOT EXISTS gender (
...
+  id smallint NOT NULL PRIMARY KEY,
+  code varchar(25) NOT NULL UNIQUE,
+  import_id bigint NOT NULL
 );
 
 /** Assessment Packages related data **/
 
 CREATE TABLE IF NOT EXISTS asmt (
-  id  bigint NOT NULL PRIMARY KEY,
+  id int NOT NULL PRIMARY KEY,
```
agorina: Interestingly this did not require Java code change. It was defined as Integer in there and worked?!

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -104,7 +106,7 @@ CREATE TABLE IF NOT EXISTS asmt_score (
 );
 
 CREATE TABLE IF NOT EXISTS claim (
-  id smallint NOT NULL PRIMARY KEY,
+  id int NOT NULL PRIMARY KEY,
```
jtreuting: Did this need to be increased?  Isn't this just all the possible claims as a lookup table?

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -212,22 +216,21 @@ CREATE TABLE IF NOT EXISTS student (
...
-  student_id bigint NOT NULL
+  student_id int NOT NULL
 );
 
 -- Note: data mart has only active groups
 CREATE TABLE IF NOT EXISTS student_group (
   id int NOT NULL PRIMARY KEY,
   name varchar(255) NOT NULL,
-  school_id mediumint NOT NULL,
+  school_id int NOT NULL,
```
jtreuting: Did you not do a smallint because you were worried about churn?  There are less than 10k schools in CA, which is the most in the nation.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -290,9 +293,9 @@ CREATE TABLE IF NOT EXISTS iab_exam_item (
   id bigint NOT NULL PRIMARY KEY,
   iab_exam_id bigint NOT NULL,
   item_id int NOT NULL,
-  score smallint,
+  score smallint NOT NULL,
```
jtreuting: Couldn't `tinyint` work here since we were told the values were -1 to 5?

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -192,14 +194,14 @@ CREATE TABLE IF NOT EXISTS item_difficulty_cuts (
 /** Data derived from the exams delivered via TRT **/
 
 CREATE TABLE IF NOT EXISTS district (
-  id mediumint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
jtreuting: same question as schools.  looks like you have a rule that if something is auto incremented then go up to int to be safe.

reporting/sql/V201702061486427077__initial_ddl.sql
agorina: There are two claims tables: one with the scored claims that are parsed from the TRT, and another one that represents claims from the tabulator. This table is the one from the tabulator.

reporting/sql/V201702061486427077__initial_ddl.sql
agorina: Yeah...32,000 seems low. 

The UNSIGNED int would be the best, but it will break our idea of being able to run the  integration tests on the DB with the data since we use negative values in there.  

Unless we are willing to change the approach for our integration tests?

reporting/sql/V201702061486427077__initial_ddl.sql
agorina: Yes, good point. Will change.

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: Pretty much. As I said above I would want to go to the UNSIGNED smallint, but we would need to change our approach to the integration tests. Maybe something to discuss with the team?

PR #51, opened 2017-05-03
Changes to support codes migration.

agorina:


reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -73,8 +73,7 @@ CREATE TABLE IF NOT EXISTS gender (
 
 CREATE TABLE IF NOT EXISTS accommodation (
   id smallint NOT NULL PRIMARY KEY,
-  code varchar(25) NOT NULL UNIQUE,
-  import_id bigint NOT NULL
```
agorina: The thought here is that accommodation is not different from all other code-tables. They change rarely, but still need to be migrated. 

To support it I have introduced a new import_content that will trigger a migration of all codes. Thoughts?

chdietz: So there is one 'code' type of import for all code tables that could enter the data warehouse from an import process.

agorina: That is correct.

PR #52, opened 2017-05-03
Added 'deleted' flag for soft deletes.

agorina:
Added 'Filipino' to ethnicity.

### DIFF
--- a/reporting/sql/V201702061486427095__initial_dml.sql
+++ b/reporting/sql/V201702061486427095__initial_dml.sql
@@ -93,7 +93,8 @@ INSERT INTO ethnicity (id, name) VALUES
   (4, 'BlackOrAfricanAmerican'),
   (5, 'White'),
   (6, 'NativeHawaiianOrOtherPacificIslander'),
-  (7, 'DemographicRaceTwoOrMoreRaces');
+  (7, 'DemographicRaceTwoOrMoreRaces'),
+  (8, 'Filipino');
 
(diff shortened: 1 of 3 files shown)


reporting/sql/V201702061486427095__initial_dml.sql
```diff
@@ -93,7 +93,8 @@ INSERT INTO ethnicity (id, name) VALUES
   (4, 'BlackOrAfricanAmerican'),
   (5, 'White'),
   (6, 'NativeHawaiianOrOtherPacificIslander'),
-  (7, 'DemographicRaceTwoOrMoreRaces');
+  (7, 'DemographicRaceTwoOrMoreRaces'),
+  (8, 'Filipino');
```
agorina: @jtreuting, I have noticed that it does not exactly matched the UI mock up. Could you please let me know if this needs to be adjusted?

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -203,6 +204,7 @@ CREATE TABLE IF NOT EXISTS school (
   natural_id varchar(40) NOT NULL UNIQUE,
```
malaffoon: Just noticed that district isn't included in all this ... that on purpose? Is the school the "master" of the district in this regard?

agorina: yes, I think we can treat school as the 'master'.

PR #54, opened 2017-05-04
School migration scrips.

agorina:
This PR includes:
1. Staging schema - first draft for the school/district only
2. Migrate scripts that will be used by the Spring Batch Migration tasks.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
 //Gradle can't call a single task more than once, because of its design, so
 //we dynamically create individual tasks for each of the available flyway commands
 //on all of our databases (dev and test)
-["reporting", "reporting-test", "warehouse", "warehouse-test"].each { schema ->
+["reporting", "reporting-test", "warehouse", "warehouse-test", "staging", "staging-test"].each { schema ->
     taskMap.keySet().each { key ->
         task("$key$schema", type: taskMap[key]) {
             group = key
--- a/staging/migrate.sql
+++ b/staging/migrate.sql
@@ -1,0 +1,98 @@
+use staging;
+-- ----------------------------------------------------------------------
+-- School / District migration
+-- ----------------------------------------------------------------------
+-- TODO: this is a temp hack/simulation that I needed to test
+TRUNCATE TABLE reporting.migrate;
+-- In the real life this will be done at the beginning of the migration
+INSERT INTO reporting.migrate (id, status, first_import_id, last_import_id)
+VALUES (11, 10, 100, 1100);
+
+-- ----------------------------------------------------------------------
+-- clean up
+-- ----------------------------------------------------------------------
+TRUNCATE TABLE staging_school;
+TRUNCATE TABLE staging_district;
+
+-- ----------------------------------------------------------------------
+-- load data into staging table
+-- ----------------------------------------------------------------------
+INSERT INTO staging_district (id, natural_id, name, migrate_id)
+  SELECT
+    wd.id,
+    wd.natural_id,
+    wd.name,
+    11 -- TODO: this id will be passed in from the previous migrate task
+  FROM warehouse.district wd
+  WHERE EXISTS(
+      SELECT id from warehouse.school ws
+      WHERE district_id = wd.id
+            AND ws.deleted = 0 -- delete will be taken care on the 'master' level
+            -- TODO: this ids will be passed in from the previous migrate task
+            AND import_id  IN ( SELECT id FROM warehouse.import WHERE id >= -1));
+
+INSERT INTO staging_school (id, natural_id, name, import_id, deleted, district_id, migrate_id)
+  SELECT
+    ws.id,
+    ws.natural_id,
+    ws.name,
+    ws.import_id,
+    ws.deleted,
+    ws.district_id,
+    11 -- TODO: this id will be passed in from the previous migrate task
+  FROM warehouse.school ws
+    JOIN warehouse.district wd ON wd.id = ws.district_id
+  WHERE
+    -- TODO: this ids will be passed in from the previous migrate task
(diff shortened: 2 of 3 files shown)


staging/migrate.sql
```diff
@@ -0,0 +1,98 @@
+use staging;
```
agorina: This scripts is a combination of the simulation of the steps done by the Spring batch before the school/district migrate, and then scrips that will be used during the migrate.

staging/migrate.sql
```diff
@@ -0,0 +1,98 @@
...
+-- ----------------------------------------------------------------------
+-- School / District migration
+-- ----------------------------------------------------------------------
+-- TODO: this is a temp hack/simulation that I needed to test
+TRUNCATE TABLE reporting.migrate;
+-- In the real life this will be done at the beginning of the migration
+INSERT INTO reporting.migrate (id, status, first_import_id, last_import_id)
+VALUES (11, 10, 100, 1100);
+
```
agorina: School/District Migration scripts start here.

staging/migrate.sql
```diff
@@ -0,0 +1,98 @@
...
+-- ----------------------------------------------------------------------
+-- clean up
+-- ----------------------------------------------------------------------
+TRUNCATE TABLE staging_school;
+TRUNCATE TABLE staging_district;
+
+-- ----------------------------------------------------------------------
+-- load data into staging table
+-- ----------------------------------------------------------------------
```
agorina: Since we cannot use CSV files for the migrate due to Aurora limitations, the above SQL will be executed in Java in 2 steps: 
1. SELECT .. FROM *warehouse*...
2. INSERT INTO *staging* ...

staging/migrate.sql
```diff
@@ -0,0 +1,98 @@
...
+    ws.deleted,
+    ws.district_id,
+    11 -- TODO: this id will be passed in from the previous migrate task
+  FROM warehouse.school ws
+    JOIN warehouse.district wd ON wd.id = ws.district_id
+  WHERE
+    -- TODO: this ids will be passed in from the previous migrate task
+    ws.import_id IN (SELECT id FROM warehouse.import WHERE id >= -1);
+
+-- -----------------------------------------------------------------------------------------
```
agorina: Starting from here, there is no more interaction with the **warehouse**. 
Now data is moved from the **staging** into **reporting**.

staging/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -0,0 +1,24 @@
+/**
+** 	Initial script for the SBAC Reporting Staging schema used during migration
+**/
+
+ALTER DATABASE staging CHARACTER SET utf8  COLLATE utf8_unicode_ci;
+
+USE staging;
+
+CREATE TABLE IF NOT EXISTS staging_district (
+  id int NOT NULL PRIMARY KEY,
```
agorina: For now we are assuming that the migrate process will migrate the data in one chunk (of import_ids) all the way through each step. This guarantees that there will be only one change to each entity within a chunk.

build.gradle
```diff
@@ -22,7 +22,7 @@ def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
 //Gradle can't call a single task more than once, because of its design, so
 //we dynamically create individual tasks for each of the available flyway commands
 //on all of our databases (dev and test)
-["reporting", "reporting-test", "warehouse", "warehouse-test"].each { schema ->
+["reporting", "reporting-test", "warehouse", "warehouse-test", "staging", "staging-test"].each { schema ->
```
agorina: @rcordeau, this was very easy to add and it works!

staging/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -0,0 +1,24 @@
...
+  natural_id varchar(40) NOT NULL,
+  name varchar(100) NOT NULL,
+  migrate_id bigint NOT NULL
```
agorina: _migrate_id_ is not being used here, but I wanted to keep it for the auditing/trouble-shooting purpose.

staging/migrate.sql
malaffoon: Do you 'need' quotes if it is a 'real' word?

staging/migrate.sql
```diff
@@ -0,0 +1,98 @@
...
+-- load data into staging table
+-- ----------------------------------------------------------------------
+INSERT INTO staging_district (id, natural_id, name, migrate_id)
+  SELECT
+    wd.id,
+    wd.natural_id,
+    wd.name,
+    11 -- TODO: this id will be passed in from the previous migrate task
+  FROM warehouse.district wd
+  WHERE EXISTS(
```
malaffoon: Do you need the `WHERE EXISTS` construct? Is that faster/better than just SELECT WHERE with a join on the import table?

staging/migrate.sql
agorina: Yes, this word always requires quotes! :)

staging/migrate.sql
agorina: ```district``` does not have import_id.
Joining with the ```school``` will give me duplicate ```districts```.
I was trying to avoid duplicates.

staging/migrate.sql
chdietz: I understand the entire flow and it looks good!

Do we want to process each root ( in the correct order ) from warehouse to stage tables before we start this step:
 - stage to reporting?

Or do all steps for each root all the way to reporting in the proper root order?

agorina: Good question. I am working on adding more 'roots' now. 
I think I want to  process each root ( in the correct order ) from warehouse to stage first, and then start the migration.

The one thing I am not sure is codes: do we want to send them through staging or write directly to the reporting? I will decided once I start writing SQL (depending on the complexity).

chdietz: I kind of like the idea of consistency, is it worth a few tables not going through stage?

agorina: You are probably right.

PR #55, opened 2017-05-04
Added Students to the migration scripts.

agorina:


staging/migrate.sql
```diff
@@ -95,4 +145,44 @@ INSERT INTO reporting.school (id, natural_id, name, district_id, import_id)
...
   FROM staging_school ss
     LEFT JOIN reporting.school rs ON rs.id = ss.id
-  WHERE rs.id IS NULL and ss.deleted = 0;
\ No newline at end of file
+  WHERE rs.id IS NULL and ss.deleted = 0;
+
+-- Student  ---------------------------------------------------------------------------------
+-- Process students update/inserts first, and then all the ethnicity at once
+
+-- TODO: do we need to wrap the below into a transaction? does it really matter?
```
agorina: Would like a feedback on the above comment.

staging/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -22,3 +22,25 @@ CREATE TABLE IF NOT EXISTS staging_school (
   deleted tinyint NOT NULL,
   migrate_id bigint NOT NULL
 );
+
+CREATE TABLE IF NOT EXISTS staging_student (
+  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
agorina: Oops, the AUTO_INCREMENT here is wrong, Removing.

staging/migrate.sql
malaffoon: Have we talked about how to handle partial failure for all this? It seems like we might want to wrap each root/section in a transaction.

staging/migrate.sql
```diff
@@ -95,4 +145,44 @@ INSERT INTO reporting.school (id, natural_id, name, district_id, import_id)
...
+UPDATE reporting.student rs
+  JOIN (SELECT id, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at,
```
chdietz: minor question, is this select in the join because you do not need all of the columns?
- some performance thing?

staging/migrate.sql
chdietz: I am thinking each step that makes up the migration of a root should be wrapped in a transaction if it is inserting/updating data.  Especially in reporting.  
- Even if it was idempotent and we ran it again we do not want 'partial' root insert/update data in reporting?
 
Some steps like populating staging tables may not require this but not sure yet.

agorina: I think we need to talk about it together. The question is "how do want to recover from the failure?". I will compile a list of my questions and set up time for us to discuss.

staging/migrate.sql
agorina: No, this is me bing silly :) I copied and pasted it from somewhere and forgot to change. Thank you for noticing and pointing it out!

PR #57, opened 2017-05-05
SQL to migrate asmt.

agorina:


staging/migrate.sql
```diff
@@ -13,7 +13,13 @@ VALUES (11, 10, 100, 1100);
 -- ----------------------------------------------------------------------
 TRUNCATE TABLE staging_school;
 TRUNCATE TABLE staging_district;
+
 TRUNCATE TABLE staging_student;
+TRUNCATE TABLE staging_student_ethnicity;
+
```
agorina: Missed this last time.

staging/migrate.sql
```diff
@@ -162,7 +235,8 @@ SET
   rs.lep_entry_at = ss.lep_entry_at,
   rs.lep_exit_at = ss.lep_exit_at,
   rs.birthday = ss.birthday,
-  rs.import_id = ss.import_id;
+  rs.import_id = ss.import_id
+WHERE ss.deleted = 0;
```
agorina: Missed that last time too.

staging/migrate.sql
```diff
@@ -184,4 +258,99 @@ INSERT INTO reporting.student (id, ssid, last_or_surname, first_name, middle_nam
...
+    LEFT JOIN reporting.asmt ra ON ra.id = sa.id
+  WHERE ra.id IS NULL and sa.deleted = 0;
+
+-- Dependent tables' changes are considered 'updates' to the asmt
+-- We need to processes them independently
+-- Note that dependent tables do not have any 'root' deletes
+-- Deleted to the depended tables are considered 'root' updates.
+
+UPDATE reporting.asmt_score ras
+  JOIN staging_asmt_score sas ON ras.asmt_id = sas.asmt_id
```
malaffoon: Does this work? Doesn't the join need more criteria? Otherwise how does it know which asmt_score to join with which?

staging/migrate.sql
```diff
@@ -184,4 +258,99 @@ INSERT INTO reporting.student (id, ssid, last_or_surname, first_name, middle_nam
...
+SET
+  ras.asmt_id     = sas.asmt_id,
```
malaffoon: If you're joining on the asmt_id you don't really need to set it.

staging/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -44,3 +48,44 @@ CREATE TABLE IF NOT EXISTS staging_student_ethnicity (
...
+  natural_id varchar(250) NOT NULL,
+  grade_id tinyint NOT NULL,
+  type_id tinyint NOT NULL,
+  subject_id tinyint NOT NULL,
+  school_year smallint NOT NULL,
+  name varchar(250),
+  label varchar(255),
+  version varchar(30),
+  import_id bigint NOT NULL,
+  deleted tinyint NOT NULL DEFAULT 0,
```
malaffoon: i wonder if having the default value might get us in trouble; it's not really needed is it?

staging/migrate.sql
agorina: Yeah, it is one to optional one relationship.

staging/migrate.sql
agorina: No, good point. Will remove.

staging/sql/V201702061486427077__initial_ddl.sql
agorina: Oops, I forgot to remove it. Copy and paste error. Thank you!

PR #58, opened 2017-05-05
Migrate codes.

agorina:


staging/migrate.sql
```diff
@@ -25,6 +42,56 @@ TRUNCATE TABLE staging_item;
 -- load data into staging table
 -- ----------------------------------------------------------------------
 
+-- Codes  ---------------------------------------------------------------
+-- Codes do not have import ids or delete flag.
+-- Since this should be a very infrequent event, we will just synchronize all the data
```
chdietz: Are you saying we will move all codes on every migration?

staging/migrate.sql
```diff
@@ -153,6 +220,318 @@ INSERT INTO staging_item ( id, claim_id, target_id, natural_id, asmt_id, math_pr
...
 
 -- -----------------------------------------------------------------------------------------
+-- MOVE FROM STAGING TO REPORTING
+-- -----------------------------------------------------------------------------------------
+
+-- Codes  ----------------------------------------------------------------------------------
+-- the following three steps should be repeated for each code table
+-- step 1: update existing codes
+-- step 2: insert new codes
+-- step 3: remove codes that do not exists in the staging/warehouse
```
chdietz: When do you see codes being moved in terms of the root order we discussed? First?
If we run each root in it's own transaction from warehouse to stage to reporting
and a code is deleted, could the reporting schema be in a bad state until all stages of the migration are complete?

chdietz: Maybe delete codes should come after all stages are complete?
I mean in stage to reporting; delete codes after other stages are complete?

staging/migrate.sql
agorina: No, the opposite. Only when there is an import with the content type 'CODES'.

staging/migrate.sql
agorina: Excellent point! You are right. It has to moved to after all is done.

staging/migrate.sql
chdietz: ok, so any import with a 'CODES' content type since the last migrate ; move all codes?

agorina: That is the idea. Do you see any issues with this approach?

chdietz: No, just wanted to make sure I understand.

PR #59, opened 2017-05-05
Migrate student groups.

agorina:
@chdietz, just to set expectations: while I am running some standalone tests before submitting a PR, I am not sure I am covering all the use cases. We need to plan to test it with all the possible use cases when we work on the migration tasks.

### DIFF
--- a/staging/migrate.sql
+++ b/staging/migrate.sql
@@ -34,6 +34,10 @@ TRUNCATE TABLE staging_district;
 TRUNCATE TABLE staging_student;
 TRUNCATE TABLE staging_student_ethnicity;
 
+TRUNCATE TABLE staging_student_group;
+TRUNCATE TABLE staging_student_group_membership;
+TRUNCATE TABLE staging_user_student_group;
+
 TRUNCATE TABLE staging_asmt;
 TRUNCATE TABLE staging_asmt_score;
 TRUNCATE TABLE staging_item;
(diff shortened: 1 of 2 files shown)


staging/migrate.sql
```diff
@@ -159,6 +163,48 @@ INSERT INTO staging_student_ethnicity (id, ethnicity_id, student_id)
...
+INSERT INTO staging_student_group_membership (student_group_id, student_id)
+  SELECT
+    wsgm.student_group_id,
+    wsgm.student_id
+  FROM warehouse.student_group_membership wsgm
+    JOIN warehouse.student_group wsg ON wsg.id= wsgm.student_group_id
+  WHERE
+    -- TODO: this ids will be passed in from the previous migrate task
+    wsg.import_id IN (SELECT id FROM warehouse.import WHERE id >= -1)
+    AND ( wsg.deleted = 0 and wsg.active = 1);  -- delete/inactive will be taken care on the 'master' level
```
agorina: Inactive student groups are not in the data mart, so we need to treat them as 'deleted'.

staging/migrate.sql
```diff
@@ -635,10 +639,67 @@ INSERT INTO reporting.student (id, ssid, last_or_surname, first_name, middle_nam
     LEFT JOIN reporting.student rs ON rs.id = ss.id
   WHERE rs.id IS NULL and ss.deleted = 0;
 
+-- TODO: clean this up:
```
agorina: I know, I was lazy. Will do it.

PR #60, opened 2017-05-07
Add spring_batch_reporting schema to gradle build and sql for ddl 

chdietz:
The ddl is for spring batch tables that it controls.
The plan is to have a spring batch schema for each migration process.
These schemas will reside on the data warehouse database server.


spring_batch_reporting/sql/V201705051494024614__spring-batch-3-0-7-RELEASE_ddl.sql
```diff
@@ -0,0 +1,102 @@
+-- Autogenerated: do not edit this file
+-- From: spring-batch 3.0.7.RELEASE spring-batch-core/src/main/resources/org/springframework/batch/core/schema-mysql.sql
+
```
agorina: I think you may need this here:
```
USE `${schemaName}`;
```

chdietz: good point!  
made the change.

PR #61, opened 2017-05-07
Migrate exams.

agorina:
This completes all the migration scripts.
As noted in prev. PR, I have tested a 'happy path' and a few other use cases, but this requires more thorough testing. 
I will play with it a little more and we should do it as we write the integration tests.


reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -155,7 +155,7 @@ CREATE TABLE IF NOT EXISTS item (
   allow_calc tinyint,
   dok_id tinyint NOT NULL,
   difficulty float NOT NULL,
-  max_points smallint UNSIGNED NOT NULL,
+  max_points tinyint UNSIGNED NOT NULL,
```
agorina: Unrelated, missed when made changes to the different int types.

reporting/data/prototype/migrate-hack.sql
```diff
@@ -1,91 +0,0 @@
-INSERT INTO reporting.district(id, name, natural_id)
```
agorina: Removed it since there is another migrate script.

reporting/sql/V201702061486427095__initial_dml.sql
```diff
@@ -17,28 +17,6 @@ INSERT INTO subject (id, name) VALUES
   (1, 'Math'),
   (2, 'ELA');
 
-INSERT INTO grade (id, code, name) VALUES
```
agorina: Removed this since it is supposed to be migrated from the warehouse.

reporting/sql/V201702061486427095__initial_dml.sql
```diff
@@ -76,96 +54,4 @@ INSERT INTO exam_claim_score_mapping (subject_claim_score_id, num) VALUES
   (13, 3),
   (14, 4);
```
agorina: Claim scores are de-normalized in the reporting mart. 
This table maps a claim id from the subject_claim_score table to a claim1-4 columns in the exam table. 
Because of this table, it makes sense to load the above two tables into the data mart.

Would it be safer to remove the above tables from the migrate scripts and assume that they must be loaded manually? Would like to know what people think.

staging/migrate.sql
```diff
@@ -3,10 +3,10 @@ use staging;
 -- MOVE FROM WAREHOUSE TO STAGING
 -- -----------------------------------------------------------------------------------------
 -- TODO: this is a temp hack/simulation that I needed to test
-TRUNCATE TABLE reporting.migrate;
+DELETE FROM reporting.migrate WHERE id = -11;
```
agorina: This makes it safe to run the script along with the real migrate (once we have one). Just wanted to keep it for testing.

staging/migrate.sql
```diff
@@ -1414,4 +1414,13 @@ DELETE rit FROM reporting.item_trait_score rit
...
+WHERE NOT EXISTS(SELECT id FROM staging_item_difficulty_cuts WHERE id = ridc.id);
+
+UPDATE reporting.migrate
+ SET
+   status = 20,
+   updated = CURRENT_TIMESTAMP,
+   message = 'manual migrate'
+WHERE id = -11;
+
+select * from reporting.migrate where id = -11;
```
agorina: Wanted to capture the run time - for testing.

reporting/sql/V201702061486427095__initial_dml.sql
malaffoon: This is more about the schema (i.e. the way the data is organized) than the data itself, so i think it belongs here.

It does raise the point that migration depends on the schema being at the same version as the code. Perhaps an early schema check and fast-fail.

agorina: Just to make sure I read your comment correctly: are you suggesting to remove these two tables from the 'migrate' process and instead use a schema version to check that it matches?

malaffoon: Yes, that is what i'm suggesting.

agorina: Makes sense. I like this approach. Thank you.

PR #62, opened 2017-05-11
updating math targets, but still waiting on correct rules to load

rcordeau:
Went through and gathered the math targets, removed duplicates where necessary, but did not include the X, XX, and XXX targets because their value depends on the 'content domain'

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/62)


agorina:
@malaffoon : yes, you are correct. Before the values were "content domain" because I mistakenly loaded them instead of the targets.

PR #64, opened 2017-05-15
updating schema names to rename *-test to  *_test

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/64)


agorina:
There are a lot of places in the migrate code that has this hardcoded.
We need to do a 'find a replace' for all of them once it is done.
Let's talk how we want to coordinate it.

rcordeau:
If you are talking about the migrate code in Ingest, I have already gone through and replaced it all and ran the integration tests against these changes.  I won't merge this PR until you see that code. We can go over it today.

agorina:
I just submitted another review that added may more..

rcordeau:
Ok, no problem. I'll get that one too. :-)

agorina:
Thank you. 
Let's coordinate the timing since Chris and I are working on adding more.

PR #67, opened 2017-05-18
Feature/accessibility codes

rcordeau:
Adding tables for accommodations codes translation 

This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/67)


warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -84,6 +84,11 @@ CREATE TABLE IF NOT EXISTS accommodation (
   code varchar(25) NOT NULL UNIQUE
 );
```
agorina: Mark asked to use Flyway with incremental updates for the warehouse. He has ETS data loaded and does not want to reprocess them if we can avoid it.

rcordeau:
Yeah for sure - I'd love to. Didn't know when people wanted to start using it incrementally.

PR #70, opened 2017-05-24
added update import id

agorina:


staging/migrate.sql
```diff
@@ -110,22 +110,22 @@ INSERT INTO staging_district (id, natural_id, name, migrate_id)
       WHERE district_id = wd.id
             AND ws.deleted = 0 -- delete will be taken care on the 'master' level
             -- TODO: this ids will be passed in from the previous migrate task
-            AND import_id  IN ( SELECT id FROM warehouse.import WHERE id >= -1));
+            AND import_id  IN ( SELECT id FROM warehouse.import WHERE id >= -1)  OR upd_import_id IN ( SELECT id FROM warehouse.import WHERE id >= -1)) ;
```
agorina: @chdietz, please note the changes to the queries since you are using them.

warehouse/sql/V201705231495560086__change_roots_import_id_ddl.sql
```diff
@@ -0,0 +1,196 @@
+/** update import id to root tables **/
+
+USE ${schemaName};
+
+ALTER TABLE asmt ADD upd_import_id bigint,
+    ADD CONSTRAINT fk__asmt__upd_import_id FOREIGN KEY (upd_import_id) REFERENCES import(id);
+UPDATE asmt SET upd_import_id = import_id;
+ALTER TABLE asmt MODIFY upd_import_id bigint NOT NULL;
+ALTER TABLE asmt ADD INDEX idx__asmt_imports_deleted (import_id, upd_import_id, deleted);
```
agorina: Not sure if this is the right one. Should I keep it or remove and wait until we have the data to test with?

staging/migrate.sql
chdietz: ok

warehouse/sql/V201705231495560086__change_roots_import_id_ddl.sql
malaffoon: As long as you have a JIRA or TODO to revisit it; we don't want unused indexes lying around.

warehouse/sql/V201705231495560086__change_roots_import_id_ddl.sql
```diff
@@ -0,0 +1,196 @@
...
+
+ALTER TABLE school ADD upd_import_id bigint NOT NULL,
```
malaffoon: `NOT NULL` shouldn't be there.

warehouse/sql/V201705231495560086__change_roots_import_id_ddl.sql
```diff
@@ -0,0 +1,196 @@
...
+    ADD CONSTRAINT fk__school__upd_import_id FOREIGN KEY (upd_import_id) REFERENCES import(id);
+UPDATE school SET upd_import_id = import_id;
+ALTER TABLE school MODIFY upd_import_id bigint NOT NULL;
+ALTER TABLE school ADD INDEX idx__asmt_imports_deleted (import_id, upd_import_id, deleted);
+
+ALTER TABLE student ADD upd_import_id bigint NOT NULL,
```
malaffoon: ditto (and now i'll stop)

warehouse/sql/V201705231495560086__change_roots_import_id_ddl.sql
agorina: I thought we have talked and agreed that it should be here?

malaffoon: if there are existing records, this will fail won't it?

agorina: Right, right...I changed it and somehow it got back in. Will fix it. Thank you!

warehouse/sql/V201705231495560086__change_roots_import_id_ddl.sql
agorina: I will add a TODO here. We have a JIRA to revisit all the TODOs.

warehouse/sql/V201705231495560086__change_roots_import_id_ddl.sql
```diff
@@ -0,0 +1,202 @@
...
+    ADD CONSTRAINT fk__iab_exam__upd_import_id FOREIGN KEY (update_import_id) REFERENCES import(id);
+UPDATE iab_exam SET update_import_id = import_id;
+ALTER TABLE iab_exam MODIFY update_import_id bigint NOT NULL;
+-- TODO: revisit this index
+ALTER TABLE iab_exam ADD INDEX idx__asmt_imports_deleted (import_id, update_import_id, deleted);
+
+ALTER TABLE exam ADD update_import_id bigint,
+    ADD CONSTRAINT fk__exam__upd_import_id FOREIGN KEY (update_import_id) REFERENCES import(id);
+UPDATE exam SET update_import_id = import_id;
+ALTER TABLE exam MODIFY update_import_id bigint NOT NULL;
```
chdietz: there is a deleted field on exam, when a delete happens where will we put the import id?

agorina: When a delete for exam happens, the update_import_id will have an id of the import request to delete and 'deleted' flag will be set to true.

chdietz: based on the `SET update_import_id = import_id` and `update_import_id bigint NOT NULL`
- when a 'root' such as exam is created we will set the import_id and the update_import_id?

agorina: That is correct. 99% of the time they will be the same.

PR #72, opened 2017-05-25
Feature/migrate translations

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/72)


reporting/sql/V201702061486427095__initial_dml.sql
```diff
@@ -22,6 +22,11 @@ INSERT INTO asmt_type (id, code, name) VALUES
   (2, 'iab', 'Interim Assessment Block'),
   (3, 'sum', 'Summative');
 
+INSERT INTO language (id, code) VALUES
+  (1, 'eng'),
+  (2, 'spa'),
+  (3, 'vie');
```
agorina: I do not think you need this here. It will be migrated, right?
We only load here data that is not migrated.

rcordeau: Ah yes, very true.

staging/migrate.sql
```diff
@@ -95,6 +98,12 @@ INSERT INTO staging_item_trait_score (id, dimension)
 INSERT INTO staging_item_difficulty_cuts (id, asmt_type_id, subject_id,grade_id, moderate_low_end, difficult_low_end)
   SELECT id, asmt_type_id, subject_id,grade_id, moderate_low_end, difficult_low_end from warehouse.item_difficulty_cuts;
 
+INSERT INTO staging_language (id, code)
```
malaffoon: Not clear on the convention of prefixing with the schema. A few hundred lines down all the tables are prefixed, should we do that here to be consistent? i.e. `INSERT INTO staging.staging_language ...`

agorina: This will be a must when it is moved into a real migrate. Otherwise it will not work. In this hack, this is really more of a 'nice' to have.

PR #73, opened 2017-05-26
Added Common Core Standards and multiple targets.

agorina:
The requirements for this change is noted on this page in "Changes to Common Core standards and Claim/Targets " section: https://confluence.fairwaytech.com/pages/viewpage.action?pageId=48824335

This review changes the schema and loads the codes.
All other changes will be in the Ingest and Migrate.


warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -190,10 +199,26 @@ CREATE TABLE IF NOT EXISTS item (
   CONSTRAINT fk__item__claim FOREIGN KEY (claim_id) REFERENCES claim(id),
   CONSTRAINT fk__item__target FOREIGN KEY (target_id) REFERENCES target(id),
   CONSTRAINT fk__item__asmt FOREIGN KEY (asmt_id) REFERENCES asmt(id),
-  CONSTRAINT fk__item__math_practice FOREIGN KEY (math_practice) REFERENCES math_practice(practice), 
+  CONSTRAINT fk__item__math_practice FOREIGN KEY (math_practice) REFERENCES math_practice(practice),
   CONSTRAINT fk__item__dok FOREIGN KEY (dok_id) REFERENCES depth_of_knowledge(id)
 );
 
+CREATE TABLE IF NOT EXISTS item_other_target (
```
agorina: Any suggestions for a better name? Or is it good enough?

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -157,11 +157,20 @@ CREATE TABLE IF NOT EXISTS subject_claim_score (
...
   claim_id smallint NOT NULL,
+  natural_id varchar(20) NOT NULL,
   code varchar(10) NOT NULL,
   description varchar(500) NOT NULL,
   CONSTRAINT fk__target__claim FOREIGN KEY (claim_id) REFERENCES claim(id)
 );
 
+CREATE TABLE IF NOT EXISTS common_core_standard (
+  id smallint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  subject_id tinyint NOT NULL,
```
malaffoon: super minor nit: by putting the subject_id before natural_id, it makes the reader (well, me) think that subject_id is somehow part of the discriminator for the entry. Which it isn't, it is just an attribute for the entry. Upshot: it makes more sense to me to have `id`, then `natural_id` then all the non-identifying attributes (ditto for target table).

warehouse/sql/V201702061486427077__initial_ddl.sql
malaffoon: good enough

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: Will change.

PR #75, opened 2017-05-30
DWR-372, make ids unique across exam and iab_exam tables

malaffoon:
I still have to do some more tests but i wanted to get feedback.

The performance of this approach is (obviously) worse than using the built-in auto-increment functionality. But it isn't too bad. From preliminary testing, a batch of 12000 inserts across two tables takes about 1800ms for built-in auto-increment and 2900ms for this scheme. It appears to be nearly invariant for table size (although i only tested up to 500k records). I did that testing with smaller test tables to avoid complications with other constraints. 

We could probably greatly improve things if we trusted that nobody would mess with the database: instead of checking every insert we could check only those inserts that have explicit `id` values.

### DIFF
--- a/warehouse/sql/V201702061486427077__initial_ddl.sql
+++ b/warehouse/sql/V201702061486427077__initial_ddl.sql
@@ -15,6 +15,59 @@ CREATE TABLE application_schema_version (
    major_version int UNIQUE NOT NULL
 );
 
+CREATE TABLE IF NOT EXISTS sequence (
+  name varchar(100) NOT NULL PRIMARY KEY,
+  increment bigint NOT NULL DEFAULT 1,
+  cur_value bigint NOT NULL DEFAULT 1
+);
+
+/**
+ * Return the next value for a named sequence and advance the sequence.
+ * The sequence must exist before calling this; for example,
+ *   INSERT INTO sequence (name) VALUES ('mysequence');
+ *   setval('mysequence', 1000);
+ *
+ * @param name of sequence
+ * @return next value in the sequence
+ */
+DROP FUNCTION IF EXISTS nextval;
+DELIMITER //
+CREATE FUNCTION nextval (seqname varchar(100))
+  RETURNS bigint NOT DETERMINISTIC
+  BEGIN
+    DECLARE curval bigint;
+    SELECT cur_value INTO curval FROM sequence WHERE name = seqname;
+    IF curval IS NOT NULL THEN
+      UPDATE sequence SET cur_value = cur_value + increment WHERE name = seqname;
+    END IF;
+    RETURN curval;
+  END //
+DELIMITER ;
+
+/**
+ * Set the next value for a named sequence. This will create the sequence if necessary.
+ *
+ * @param name sequence name
+ * @param newval sequence next value
+ * @return next value in the sequence (same as newval)
+ */
+DROP FUNCTION IF EXISTS setval;
+DELIMITER //
+CREATE FUNCTION setval (seqname varchar(100), newval bigint)
+  RETURNS bigint NOT DETERMINISTIC
+  BEGIN
+    DECLARE curval bigint;
+    SELECT cur_value INTO curval FROM sequence WHERE name = seqname;
+    IF curval IS NOT NULL THEN
+      UPDATE sequence SET cur_value = newval WHERE name = seqname;
+    ELSE
+      INSERT INTO sequence (name, cur_value) VALUES (seqname, newval);
+    END IF;
+    RETURN newval;
+  END //
+DELIMITER ;
+
+
 /** Import **/
 
 CREATE TABLE IF NOT EXISTS import_content (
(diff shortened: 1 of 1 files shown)


warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -15,6 +15,59 @@ CREATE TABLE application_schema_version (
...
+/**
+ * Return the next value for a named sequence and advance the sequence.
+ * The sequence must exist before calling this; for example,
+ *   INSERT INTO sequence (name) VALUES ('mysequence');
+ *   setval('mysequence', 1000);
+ *
+ * @param name of sequence
+ * @return next value in the sequence
+ */
+DROP FUNCTION IF EXISTS nextval;
```
malaffoon: I added this as a general capability, similar to postgresql.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -633,3 +686,56 @@ CREATE PROCEDURE school_upsert(IN  p_district_name       VARCHAR(100),
     END IF;
   END; //
 DELIMITER ;
+
+
+/** Cross-table auto-increment for exam and iab_exam **/
+
+INSERT IGNORE INTO sequence (name) VALUES ('exam_id');
```
malaffoon: create the sequence we'll be using

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -633,3 +686,56 @@ CREATE PROCEDURE school_upsert(IN  p_district_name       VARCHAR(100),
...
+DROP PROCEDURE IF EXISTS exam_insert_helper;
+DELIMITER //
+CREATE PROCEDURE exam_insert_helper(INOUT new_id bigint)
+  BEGIN
+    DECLARE tmpid bigint;
+    DECLARE msg varchar(120);
+    IF new_id IS NULL OR new_id = 0 THEN
+      SET new_id = nextval('exam_id');
+    END IF;
+    IF EXISTS(SELECT id FROM exam WHERE id = new_id) OR EXISTS(SELECT id FROM iab_exam WHERE id = new_id) THEN
```
malaffoon: This is the check i was talking about for performance. We could make this conditional on new_id having been set by the caller. But that requires trusting that nobody messed with the sequence by setting the nextval below any value already in the tables.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -15,6 +15,59 @@ CREATE TABLE application_schema_version (
...
+DELIMITER //
+CREATE FUNCTION nextval (seqname varchar(100))
+  RETURNS bigint NOT DETERMINISTIC
+  BEGIN
+    DECLARE curval bigint;
+    SELECT cur_value INTO curval FROM sequence WHERE name = seqname;
+    IF curval IS NOT NULL THEN
+      UPDATE sequence SET cur_value = cur_value + increment WHERE name = seqname;
+    END IF;
+    RETURN curval;
```
agorina: I think two concurrent requests with a certain timing may end up with the same id here.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -15,6 +15,59 @@ CREATE TABLE application_schema_version (
    major_version int UNIQUE NOT NULL
 );
 
+CREATE TABLE IF NOT EXISTS sequence (
+  name varchar(100) NOT NULL PRIMARY KEY,
+  increment bigint NOT NULL DEFAULT 1,
+  cur_value bigint NOT NULL DEFAULT 1
```
agorina: I understand what you are trying to achieve, but I probably would feel better if cur_value was an AUTO_INCREMENT column. From all what I read there is a lot of optimization in InnoDb and AUTO_INCREMENT column.

agorina:
I am just thinking loud here. I am sure you have done tons of research and may have see this one, but just in case:  I like the last suggestion in this link: https://www.quora.com/Should-I-share-an-auto_incremented-id-across-tables 
I know it is not generic, but it feels right to me if we want to have the unique id across two tables.

warehouse/sql/V201702061486427077__initial_ddl.sql
malaffoon: I had the same concern, i just don't know enough psql to know how to code around it.

warehouse/sql/V201702061486427077__initial_ddl.sql
malaffoon: And how do i do that without adding rows?

There is a completely different approach that uses that idea: you have a master exam table that has the auto-increment, and then the iab and non-iab tables "extend" the master. Classic ORM approach. I could do that if you want ...

malaffoon:
We have decided to merge exam and iab_exam so this is not necessary.

PR #76, opened 2017-05-30
Fix migrate typos.

agorina:


staging/migrate.sql
```diff
@@ -51,7 +51,7 @@ TRUNCATE TABLE staging_asmt;
 TRUNCATE TABLE staging_asmt_score;
 TRUNCATE TABLE staging_item;
 
-TRUNCATE TABLE staging_acommodation_translation;
```
agorina: This was a typo.

staging/migrate.sql
```diff
@@ -646,22 +646,22 @@ INSERT INTO reporting.target ( id, claim_id, code, description)
   WHERE rt.id IS NULL;
 
 -- ------------ Common Core Standard -------------------------------------------------------------------------
-UPDATE reporting.common_core_standard rc
-    JOIN staging.staging_common_core_standard sccs ON sccs.id = rccs.id
- SET
-    rccs.subject_id = sccs.subject_id,
-    rccs.natural_id = sccs.natural_id,
-    rccs.description = sccs.description;
+UPDATE reporting.common_core_standard rccs
```
agorina: And this was wrong.

staging/migrate.sql
chdietz: spelling error, why didn't it fail ?

agorina: It fails when I run this script. I am assuming that this is just a typo in here, not in the migrate code?

PR #78, opened 2017-06-01
Corrections to item and asmt cut points.

agorina:
I was wrong about item correction. But cut points still need to be adjusted.

### DIFF
--- a/warehouse/sql/V201702061486427077__initial_ddl.sql
+++ b/warehouse/sql/V201702061486427077__initial_ddl.sql
@@ -127,9 +127,9 @@ ALTER TABLE asmt ADD INDEX idx__asmt_imports_deleted (import_id, update_import_i
 
 CREATE TABLE IF NOT EXISTS asmt_score (
   asmt_id int NOT NULL PRIMARY KEY,
-  cut_point_1 float NOT NULL,
+  cut_point_1 float,
   cut_point_2 float NOT NULL,
-  cut_point_3 float NOT NULL,
+  cut_point_3 float,
   min_score float NOT NULL,
   max_score float NOT NULL,
   CONSTRAINT fk__asmt_score__asmt FOREIGN KEY (asmt_id) REFERENCES asmt(id)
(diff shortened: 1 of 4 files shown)


warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -189,7 +189,7 @@ CREATE TABLE IF NOT EXISTS item (
   id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
   claim_id smallint, -- TODO: change this to NOT NULL when we do not need a hack code that creates assmt
   target_id smallint, -- TODO: change this to NOT NULL  when we do not need a hack code that creates assmt
-  natural_id varchar(40) NOT NULL,
+  natural_id varchar(40) NOT NULL UNIQUE,
```
agorina: I really hope this is right. @jtreuting , could you please comment on it?

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -127,9 +127,9 @@ ALTER TABLE asmt ADD INDEX idx__asmt_imports_deleted (import_id, update_import_i
 
 CREATE TABLE IF NOT EXISTS asmt_score (
   asmt_id int NOT NULL PRIMARY KEY,
-  cut_point_1 float NOT NULL,
+  cut_point_1 float,
   cut_point_2 float NOT NULL,
-  cut_point_3 float NOT NULL,
```
agorina: The asmt package loader does not load 1 and 3 for IAB.

malaffoon: Why, do we really want to throw away data? Just because they aren't used doesn't mean they aren't real data.

agorina: The code does ignore these values for IABs since I did not think they add any value. I thought that it would be confusing to have them if they are meaningless. I guess maybe I do not understand their value. Could you please clarify?

malaffoon: The cut-points for any assessment are mathematically meaningful. They represent specific likelihoods of the students real capability. The problem with IABs is that, although meaningful, the data is unlikely to be statistically significant, making conclusions drawn from those values less justified. However, mathematically, the values are still "real", they are part of the definition of the assessment characteristics.

PR #81, opened 2017-06-04
Modify migrate update date to be updated.

agorina:


reporting/sql/V201706011496359199__migrate_update_timestamp_ddl.sql
```diff
@@ -0,0 +1,7 @@
+/**
+* change migrate update column to be refreshed on update
+**/
+
+USE ${schemaName};
+
+ALTER TABLE migrate MODIFY COLUMN updated timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6);
```
agorina: I have noticed that updated date was not updated.

PR #84, opened 2017-06-06
Feature/create artifacts

rcordeau:
Adding in ability to publish schema as a zip archive

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/84)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,17 @@ buildscript {
 
 plugins {
     id 'org.flywaydb.flyway' version '4.2.0'
+    id 'com.jfrog.artifactory' version '4.4.18'
+}
+
+apply plugin: 'distribution'
+apply plugin: 'maven-publish'
+
+group = 'org.opentestsystem.rdw.schema'
+version = '0.0.1-SNAPSHOT'
+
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -11,6 +11,16 @@ plugins {
     id 'org.flywaydb.flyway' version '4.2.0'
 }
 
+apply plugin: 'distribution'
+apply plugin: 'maven-publish'
+
+group = 'org.opentestsystem.rdw.schema'
+version = '0.0.1-SNAPSHOT'
+
+if (project.hasProperty("buildNumber")) {
```
malaffoon: Does this have to be conditional? Won't all our projects have this?

rcordeau: It is conditional because we don't always specify the build number on the command line. And the property doesn't exist until it is specified on the command line, unless you'd like it specified in a gradle.properties file instead. The buildNumber (-PbuildNumber) is only specified when running on the CI server, that way it will default to SNAPSHOT.

malaffoon: Yup, figured that out after you answer the other question. I wasn't making the connection between project.hasProperty() and command-line.

PR #85, opened 2017-06-07
adding an install task that will publish to mavenLocal. Useful for de…

rcordeau:
…veloping locally.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/85)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,6 +85,9 @@ publishing {
     }
 }
 
+//Install task for installing (publishing) to local .m2
+task install(dependsOn: publishToMavenLocal)
+
 
 //For ci use only
 artifactory {


build.gradle
```diff
@@ -85,6 +85,9 @@ publishing {
     }
 }
 
+//Install task for installing (publishing) to local .m2
+task install(dependsOn: publishToMavenLocal)
```
rcordeau: Yeah, we could not have this and just use "./gradlew publishToMavenLocal", but "install" is less typing.

malaffoon: 👍

agorina: yeah!!!! no more submodules?

rcordeau: Nope.

chdietz: nice!

PR #88, opened 2017-06-13
Part of the performance improvement for migrate.

agorina:
Changes to the migrate code are to follow.


staging/migrate.sql
```diff
@@ -1,1263 +0,0 @@
-use staging;
```
agorina: Unrelated, I removed it.

PR #90, opened 2017-06-14
adding two new tasks for data mart related schemas

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/90)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ if (project.hasProperty("buildNumber")) {
 }
 
 //Setting up an easy mapping of simple name to task object
-def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
+def flywayActions = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
                clean: org.flywaydb.gradle.task.FlywayCleanTask,
                info: org.flywaydb.gradle.task.FlywayInfoTask,
                validate: org.flywaydb.gradle.task.FlywayValidateTask,
@@ -37,15 +37,15 @@ def warehouseSchemas = ["warehouse"]
 
 //Gradle can't call a single task more than once, because of its design, so
 //we dynamically create individual tasks for each of the available flyway commands
-//on all of our databases (dev and test)
+//on all of our databases (dev and test). Task names end up like migrateStaging and cleanWarehouse_test.
 (dataMartSchemas + warehouseSchemas).each { schema ->
-    taskMap.keySet().each { key ->
+    flywayActions.keySet().each { action ->
         ["", "_test"].each { type ->
-            task("$key$schema$type", type: taskMap[key]) {
-                group = "$key$type"
-                url = 'jdbc:mysql://localhost:3306/'
-                user = 'root'
-                password = ''
+            task(action + schema.capitalize() + type, type: flywayActions[action]) {
+                group = "$action$type"
+                url = database_url
+                user = database_user
+                password = database_password
                 driver = 'com.mysql.jdbc.Driver'
                 schemas = ["$schema$type"]
                 locations = ['filesystem:' + rootDir.path + "/$schema/sql"]
@@ -55,8 +55,8 @@ def warehouseSchemas = ["warehouse"]
     }
 }
 
+def ts = project.getTasks()
 ["clean", "migrate"].each { allTask ->
-    def ts = project.getTasks()
     ["","_test"].each { type ->
         task("${allTask}All$type", dependsOn: [ts.findAll { it.property("group") == "$allTask$type" }]) {
             group = "$allTask$type"
--- a/build.gradle
+++ b/build.gradle
@@ -30,10 +30,13 @@ def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
                baseline: org.flywaydb.gradle.task.FlywayBaselineTask,
                repair: org.flywaydb.gradle.task.FlywayRepairTask]
 
+def dataMartSchemas = ["reporting", "reporting_test", "staging", "staging_test", "spring_batch_reporting", "spring_batch_reporting_test"]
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -50,16 +53,26 @@ def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
 }
 
 ["clean", "migrate"].each { allTask ->
-    task("${allTask}All", dependsOn: [project.getTasks().findAll{x-> !(x.name.contains("_test")) && x.property("group") == allTask}]){
+    def ts = project.getTasks()
+    task("${allTask}DataMart", dependsOn: [ts.findAll{x -> !(x.name.contains('warehouse')) &&  !(x.name.contains("_test")) && x.property("group") == allTask}]){
```
malaffoon: The dependsOn bothers me a bit: is there a way you can use the schema lists instead of relying on "warehouse" in the name? It doesn't take much to imagine adding another schema in the warehouse set of schemas. Perhaps add an attribute to the tasks and use it in the filter?

chdietz: Mark may have a good idea, i tested it first and noticed the same issue in the code after running into issues noted below.

rcordeau: Yeah, it was late and I was being quite lazy. I'll fix it up.

build.gradle
```diff
@@ -30,33 +32,40 @@ def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
...
 
 ["clean", "migrate"].each { allTask ->
-    task("${allTask}All", dependsOn: [project.getTasks().findAll{x-> !(x.name.contains("_test")) && x.property("group") == allTask}]){
-        group=allTask
-        description="Custom task to " + allTask.capitalize() + " all objects in the configured production/development schemas."
-    }
-    task("${allTask}All-test", dependsOn: [project.getTasks().findAll{x-> x.name.contains("_test") && x.property("group") == allTask}]){
-        group=allTask
-        description="Custom task to " + allTask.capitalize() + " all objects in the configured test schemas."
+    def ts = project.getTasks()
```
rcordeau: A bit more convoluted, but basically factors out a lot of other code. Removed dependency on checking for the "_test" bare string in task names for cleanAll and migrateAll, and got more "groovy" in filtering the data mart tasks against the data mart schema list defined earlier in the code.

build.gradle
```diff
@@ -30,33 +32,40 @@ def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
...
+    ["","_test"].each { type ->
+        task("${allTask}All$type", dependsOn: [ts.findAll { it.property("group") == "$allTask$type" }]) {
+            group = "$allTask$type"
+            description = "Custom task to " + allTask.capitalize() + MessageFormat.format(" all objects in the configured {0} schemas.", type ? "test" : "production/development")
+        }
+        task("${allTask}DataMart$type", dependsOn: [ts.findAll{x-> x.name in dataMartSchemas.collect {"$allTask$it$type".toString()}}]) {
+            group = "$allTask$type"
+            description = "Custom task to " + allTask.capitalize() +  MessageFormat.format(" all objects in the configured {0} data mart schemas.", type ? "test" : "production/development")
```
rcordeau: empty strings ("") evaluate to false in groovy.

build.gradle
```diff
@@ -30,33 +32,40 @@ def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
...
 
+def dataMartSchemas = ["reporting", "staging", "spring_batch_reporting"]
+def warehouseSchemas = ["warehouse"]
+
 //Gradle can't call a single task more than once, because of its design, so
 //we dynamically create individual tasks for each of the available flyway commands
 //on all of our databases (dev and test)
-["reporting", "reporting_test", "warehouse", "warehouse_test", "staging", "staging_test", "spring_batch_reporting", "spring_batch_reporting_test"].each { schema ->
+(dataMartSchemas + warehouseSchemas).each { schema ->
     taskMap.keySet().each { key ->
```
malaffoon: i know it is more obvious when viewing the whole file and not just these diffs, but perhaps rename `taskMap` to `flywayActions` or something.

build.gradle
```diff
@@ -30,33 +32,40 @@ def taskMap = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
                baseline: org.flywaydb.gradle.task.FlywayBaselineTask,
                repair: org.flywaydb.gradle.task.FlywayRepairTask]
 
+def dataMartSchemas = ["reporting", "staging", "spring_batch_reporting"]
+def warehouseSchemas = ["warehouse"]
+
 //Gradle can't call a single task more than once, because of its design, so
 //we dynamically create individual tasks for each of the available flyway commands
 //on all of our databases (dev and test)
```
malaffoon: Perhaps a continuation of this comment, something like `... and test). Task names end up like migratestaging and cleanwarehouse_test.

build.gradle
malaffoon: i think this line can be before line 58? no biggie but, the way it is, i spent a few minutes figuring whether the allTasks depend on other allTasks.

build.gradle
rcordeau: Sounds good.

build.gradle
rcordeau: I can move it. It just affects the scope.

build.gradle
rcordeau: I’m wondering if it would be more clear if the commands were more like migrate_staging or clean_warehouse_test

malaffoon: it would be clearer when reading them, yes

rcordeau: Well, the gradle convention seems to be camelCase, so I'll keep with that and update the task name to be more camelCase.

PR #94, opened 2017-06-19
Added an explicit school index.

agorina:
This is a fix to the rollback I had to do late Friday.

### DIFF
--- a/reporting/sql/V201706191497893750__added_school_index_to_exam_ddl.sql
+++ b/reporting/sql/V201706191497893750__added_school_index_to_exam_ddl.sql
@@ -1,0 +1,14 @@
+/**
+* MySQL treats FK this way:
+* In the referencing table, there must be an index where the foreign key columns are listed as the first columns in the same order.
+* Such an index is created on the referencing table automatically if it does not exist.
+* This index is silently dropped later, if you create another index that can be used to enforce the foreign key constraint.
+*
+* When restoring a DB from a back up, MySQL does not see an automatically created FK index as such and treats it as a user defined.
+* So when running this on the restored DB, you will end up with duplicate indexes.
+**/
+
+ALTER TABLE exam DROP FOREIGN KEY fk__exam__school;
+
+CREATE INDEX idx__exam__school ON exam (school_id);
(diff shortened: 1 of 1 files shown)


reporting/sql/V201706191497893750__added_school_index_to_exam_ddl.sql
```diff
@@ -0,0 +1,14 @@
...
+* MySQL treats FK this way:
+* In the referencing table, there must be an index where the foreign key columns are listed as the first columns in the same order.
+* Such an index is created on the referencing table automatically if it does not exist.
+* This index is silently dropped later, if you create another index that can be used to enforce the foreign key constraint.
+*
+* When restoring a DB from a back up, MySQL does not see an automatically created FK index as such and treats it as a user defined.
+* So when running this on the restored DB, you will end up with duplicate indexes.
+**/
+
+ALTER TABLE exam DROP FOREIGN KEY fk__exam__school;
```
agorina: This does not drop the index. So when the DB is restored from the back up, the autocreated fk__exam_school is defined there, but it does not think/know that it is an autocreated one.

It feels like our initial syntax for the FK creation could be improved by first creating an explicit indexes and then defining FK. Then we would not have this issue.

malaffoon: so yeah, let's do that

agorina: Do you want to do it now? 
Or when time comes to merge the incremental updates?

agorina: https://jira.fairwaytech.com/browse/DWR-459

chdietz: Forgot to comment on this, so at some point we will modify ddl to create all FK in the same way?

agorina: Yes, I have created a JIRA:https://jira.fairwaytech.com/browse/DWR-459.

PR #95, opened 2017-06-20
Feature/dwr 413 denormalize student info

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/95)

### DIFF
--- a/reporting/sql/V201706151497570237__denormalize_asmt_scores_exam_student_info_codes_ddl.sql
+++ b/reporting/sql/V201706151497570237__denormalize_asmt_scores_exam_student_info_codes_ddl.sql
@@ -1,0 +1,54 @@
+/**
+* DWR-413 Reporting datamart: denormalize code fields for student and student_ethnicity
+**/
+
+USE ${schemaName};
+
+-- add gender_code
+ALTER TABLE student ADD COLUMN gender_code varchar(255);
+update student s
+  join gender g on s.gender_id = g.id
+ set s.gender_code = g.code;
+
+-- add ethnicity code
+ALTER TABLE student_ethnicity ADD COLUMN ethnicity_code varchar(255);
+update student_ethnicity se
+   join ethnicity e on e.id = se.ethnicity_id
+  set se.ethnicity_code = e.code;
+
+ALTER TABLE student MODIFY COLUMN gender_code varchar(255) NOT NULL;
+ALTER TABLE student_ethnicity MODIFY COLUMN ethnicity_code varchar(255) NOT NULL;
+
+-- add asmt_score columns to asmt
+ALTER TABLE asmt ADD COLUMN cut_point_1 smallint;
+ALTER TABLE asmt ADD COLUMN cut_point_2 smallint;
+ALTER TABLE asmt ADD COLUMN cut_point_3 smallint;
+ALTER TABLE asmt ADD COLUMN min_score smallint;
+ALTER TABLE asmt ADD COLUMN max_score smallint;
+
+UPDATE asmt a
+	JOIN asmt_score ascore ON ascore.asmt_id = a.id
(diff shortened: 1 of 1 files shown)


reporting/sql/V201706151497570237__denormalize_asmt_scores_exam_student_info_codes_ddl.sql
```diff
@@ -0,0 +1,54 @@
+/**
+* DWR-413 Reporting datamart: denormalize code fields for student and student_ethnicity
+**/
+
+USE ${schemaName};
+
+-- add gender_code
+ALTER TABLE student ADD COLUMN gender_code varchar(255);
```
malaffoon: I guess you're just taking this from the existing schema, but 255 seems silly for gender and ethnicity code. Not suggesting a change, just whining.

reporting/sql/V201706151497570237__denormalize_asmt_scores_exam_student_info_codes_ddl.sql
```diff
@@ -0,0 +1,54 @@
...
+
+-- add ethnicity code
+ALTER TABLE student_ethnicity ADD COLUMN ethnicity_code varchar(255);
+update student_ethnicity se
+   join ethnicity e on e.id = se.ethnicity_id
+  set se.ethnicity_code = e.code;
+
+ALTER TABLE student MODIFY COLUMN gender_code varchar(255) NOT NULL;
+ALTER TABLE student_ethnicity MODIFY COLUMN ethnicity_code varchar(255) NOT NULL;
+
```
malaffoon: We're keeping the gender_id and ethnicity_id columns?

reporting/sql/V201706151497570237__denormalize_asmt_scores_exam_student_info_codes_ddl.sql
rcordeau: Yeah, I had the same whine.

reporting/sql/V201706151497570237__denormalize_asmt_scores_exam_student_info_codes_ddl.sql
agorina: Yes, we have decided to keep both. Do you see any problems with this approach?

rcordeau: Yeah I talked to Alla and the idea was that it wasn't a big deal to keep them in.

agorina: @esotrope, I think you had a reason why you wanted both. Could you please comment?

reporting/sql/V201706151497570237__denormalize_asmt_scores_exam_student_info_codes_ddl.sql
agorina: whiners :)

reporting/sql/V201706151497570237__denormalize_asmt_scores_exam_student_info_codes_ddl.sql
esotrope: Don't think there is a reason to keep both but i know we decided to.
For all ID columns that reference an entity and eventually just resolve to a code we won't ever use the ID column after denormalizing it

agorina: Oh, I thought you had a better reason :)

I still think that it would be easier to troubleshoot this way, and it is really not a big overhead to keep them. But if people object it, we can remove them.

malaffoon: i don't care either way, just asking in case it was an oversight vs. an intentional omission.

PR #96, opened 2017-06-21
DWR-448 Documentation: create tips and tricks document

agorina:


RunBook.md
```diff
@@ -0,0 +1,91 @@
...
+The ```migrate``` table is the main control table for the migrate. It has the last successfully imported import id from the warehouse. 
+
+The migrate process is managed by the “migrate-reporting” service.  
+
+### Maintenance Guidelines
+**Note 1:** 
+>Only data updates are supported, not structural table changes.
+
+**Note 2:**  
+>The best practice is to use the Import API to ingest the data. In a rare case you need to update the data manually, follow the instructions below.
```
chdietz: In a rare case that requires a manual data update,

RunBook.md
```diff
@@ -0,0 +1,91 @@
...
+mysql>INSERT INTO import(status, content, contentType, digest) VALUES (1, 3, 'initial load', 'initial load');
+```
+- The migrate will pick this up. It will migrate all tables from this category.
+
+#### Modify one main table and any of its children
+- Create an import id to associate with your changes. Use an appropriate content type:
+```sql
+mysql> USE warehouse;
+mysql> INSERT INTO import (status, content, contentType, digest, creator) VALUES (0, 5, 'text/plain', left(uuid(), 8), 'dwtest@example.com');
+mysql> SELECT LAST_INSERT_ID() into @IMPORT_ID;
```
chdietz: If there is a ingest in process could you get the wrong id?
Maybe pause ingest and migrate then run this or know your digest and query from that?

RunBook.md
```diff
@@ -0,0 +1,91 @@
...
+PACKAGES | **asmt**, asmt_score, item, item_common_core_standard, item_other_target | Ingested using the Import Service API and the output from the tabulator.
+ORGANIZATION | **school**, district | Ingested using the ART Organizations extract.
+GROUPS | **student_group**, student, student_group_membership, user_student_group | Uploaded via Group Management API. For ```student``` only student SSID is available from this source.
+EXAM | **exam**, student, exam_student, exam_item, exam_available_accommodation, exam_claim_score | Ingested from TRTs.
+
+#### Import table and Import ID
+
+```import``` table is the main control table for importing the data. In order for the system to function properly any data modifications **must** be logged into the ```import``` table.
+Each content type has a main table associated with it. It is listed as the first table on the table list above. These main tables have 3 common columns:
```
chdietz: For terminology, do we want to use 'main' or maybe 'root'?  Once we publish terminology it's hard to change.

agorina: I think 'main' is more appropriate. I was struggling with a 'root' having 'children'.

RunBook.md
agorina: I do not think you can get the wrong ID this way. Based on my research this should give you the last ID you inserted in your session.

RunBook.md
agorina: Will rephrase. Thank you.

RunBook.md
chdietz: ok, cool, it looks like it is per connection.  Good to know.

PR #97, opened 2017-06-27
DWR-459 'Data Intervention'

agorina:
Includes the following:
1. Merged all incremental schemas
2. Modify initial DDL to explicitly create indexes for FK.  If you are curious why, you can find more here https://dev.mysql.com/doc/refman/5.6/en/create-table-foreign-keys.html.   Below is an excerpt from the referenced page. We ran into issues with the *silently dropped* indexes and want to control them better.
> "...MySQL requires indexes on foreign keys and referenced keys so that foreign key checks can be fast and not require a table scan. In the referencing table, there must be an index where the foreign key columns are listed as the first columns in the same order. **Such an index is created on the referencing table automatically if it does not exist. This index might be silently dropped later, if you create another index that can be used to enforce the foreign key constraint** ..."
3. Schema changes for DWR-427: Student Group data warehouse model changes. All `student` attributes but SSID are optional.
4. Schema changes for DWR-480: Added `opptId` to the `warehouse`
6. Schema changes for DWR-464: Added  `position` to `item`.

I will create a new DB dump once it is approved.


reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -11,10 +11,6 @@ ALTER DATABASE ${schemaName} CHARACTER SET utf8 COLLATE utf8_unicode_ci;
 
 USE ${schemaName};
 
-CREATE TABLE application_schema_version (
-   major_version int UNIQUE NOT NULL
-);
-
```
agorina: Unrelated, removed since it is not used.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -64,40 +60,34 @@ CREATE TABLE IF NOT EXISTS administration_condition (
...
   id tinyint NOT NULL PRIMARY KEY,
-  code varchar(255) NOT NULL UNIQUE
+  code varchar(120) NOT NULL UNIQUE
 );
 
 CREATE TABLE IF NOT EXISTS gender (
   id tinyint NOT NULL PRIMARY KEY,
-  code varchar(255) NOT NULL UNIQUE
+  code varchar(80) NOT NULL UNIQUE
 );
```
agorina: Since some people complained about the large size here and since we are preloading this data anyway, I have decreased it (here and above).

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -172,15 +182,24 @@ CREATE TABLE IF NOT EXISTS math_practice (
...
-  difficulty float NOT NULL,
+  difficulty_code varchar(1) NOT NULL,
   max_points tinyint UNSIGNED NOT NULL,
+  position tinyint,
+  claim_code varchar(10) NOT NULL,
+  target_code varchar(10) NOT NULL,
+  common_core_standard_ids varchar(200),
+  INDEX idx__item__claim (claim_id),
+  INDEX idx__item__target (target_id),
+  UNIQUE INDEX idx__item__asmt_natural_id (asmt_id, natural_id),
```
malaffoon: nit: suggest putting UNIQUE INDEX's first; unless there is another order here that i'm not seeing. It could be you did it to match the column order, in which case i'd suggest moving natural_id above claim_id and target_id.

reporting/sql/V201706011496359191__add_claim_codes_to_asmt_ddl.sql
```diff
@@ -1,33 +0,0 @@
-/**
-** add claim 1-4 codes from the exam table to the asmt
```
chdietz: so, you consolidated the current flyway scripts?

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -191,78 +210,71 @@ CREATE TABLE IF NOT EXISTS item (
 CREATE TABLE IF NOT EXISTS item_other_target (
   item_id int NOT NULL,
   target_id smallint NOT NULL,
+  UNIQUE INDEX idx__item_target (item_id, target_id),
+  INDEX idx__item_taget__target (target_id),
```
malaffoon: typo

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -191,78 +210,71 @@ CREATE TABLE IF NOT EXISTS item (
...
   CONSTRAINT fk__item_target__item FOREIGN KEY (item_id) REFERENCES item(id),
-  CONSTRAINT fk__item_taget__target FOREIGN KEY (target_id) REFERENCES target(id),
-  CONSTRAINT uk__item_target UNIQUE INDEX (item_id, target_id)
+  CONSTRAINT fk__item_taget__target FOREIGN KEY (target_id) REFERENCES target(id)
```
malaffoon: same typo

reporting/sql/V201702061486427077__initial_ddl.sql
malaffoon: nit: slight inconsistency in naming UNIQUE INDEX's; elsewhere you don't include the column names in the index name, so this would be just `idx__item`.

reporting/sql/V201702061486427077__initial_ddl.sql
agorina: Ughhhh...will fix.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -191,78 +210,71 @@ CREATE TABLE IF NOT EXISTS item (
...
-  name varchar(100) NOT NULL
-
+  natural_id varchar(40) NOT NULL,
+  name varchar(100) NOT NULL,
+  UNIQUE INDEX idx__district__natural_id (natural_id)
 );
 
 CREATE TABLE IF NOT EXISTS school (
-  id int NOT NULL PRIMARY KEY,
+  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
malaffoon: Really? An AUTO_INCREMENT in the reporting data mart?

agorina: Good catch, it is an error. Will remove.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -191,78 +210,71 @@ CREATE TABLE IF NOT EXISTS item (
...
-
 /** Student Groups */
 
 CREATE TABLE IF NOT EXISTS student (
   id int NOT NULL PRIMARY KEY,
-  ssid varchar(65) NOT NULL UNIQUE,
-  last_or_surname varchar(60) NOT NULL,
-  first_name varchar(60) NOT NULL,
+  ssid varchar(65) NOT NULL,
+  last_or_surname varchar(60),
```
malaffoon: unrelated to your changes but this, combined with @esotrope's recent review ...

Why is it `last_or_surname` but not `first_or_givenname`? Seems like we should just make it `last_name` and give in to our western hemisphere bias.

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -191,78 +210,71 @@ CREATE TABLE IF NOT EXISTS item (
...
+  first_name varchar(60),
   middle_name varchar(60),
-  gender_id tinyint NOT NULL,
+  gender_id tinyint,
```
malaffoon: nit: could have that `-- intentionally not constrained...` comment?

reporting/sql/V201702061486427077__initial_ddl.sql
agorina: The name was copied from SBAC spec.

chdietz: TRT data model calls it LastOrSurname

reporting/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -11,10 +11,6 @@ ALTER DATABASE ${schemaName} CHARACTER SET utf8 COLLATE utf8_unicode_ci;
```
malaffoon: Perhaps a blurb at the top of both the *initial_ddl.sql scripts about the FK / INDEX behavior of MySQL?

reporting/sql/V201706191497893750__added_school_index_to_exam_ddl.sql
```diff
@@ -1,14 +0,0 @@
-/**
-* MySQL treats FK this way:
```
malaffoon: Yeah, this ... in the main script.

reporting/sql/V201702061486427077__initial_ddl.sql
agorina: Good point, will add.

reporting/sql/V201706011496359191__add_claim_codes_to_asmt_ddl.sql
agorina: yes

staging/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -123,22 +110,23 @@ CREATE TABLE IF NOT EXISTS staging_school (
...
   import_id bigint NOT NULL,
   deleted tinyint NOT NULL,
   migrate_id bigint NOT NULL
 );
 
 CREATE TABLE IF NOT EXISTS staging_student_ethnicity (
   ethnicity_id tinyint NOT NULL,
-  student_id int NOT NULL
+  student_id int NOT NULL,
+  UNIQUE INDEX idx__staging_student_ethnicity__student_ethnicity (student_id, ethnicity_id)
```
malaffoon: Should there be any constraints in the staging tables? The `UNIQUE` part not the `INDEX` part. Is there a performance benefit or optimizer hint or something by making it `UNIQUE`?

agorina: It is intentional. Discussed offline, will add comments explaining it in the script.

chdietz: considering the batch size for staging if this is for performance would the whole table fit in memory?
if it is for validity then i thought the rule is that staging relies on warehouse, and it has to migrate to reporting and it has constraints.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -240,45 +263,47 @@ CREATE TABLE IF NOT EXISTS item_difficulty_cuts (
...
-
 /** Student Groups */
 
 CREATE TABLE IF NOT EXISTS student (
   id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
-  ssid varchar(65) NOT NULL UNIQUE,
-  last_or_surname varchar(60) NOT NULL,
-  first_name varchar(60) NOT NULL,
+  ssid varchar(65) NOT NULL,
+  last_or_surname varchar(60),
```
chdietz: Wondering how we arrived at 60, guess it is ok, but TRDD has the name fields at 35

staging/sql/V201702061486427077__initial_ddl.sql
agorina: This was done in this PR: https://github.com/SmarterApp/RDW_Schema/pull/88/files and it is a part of the optimization when the data get moved from warehouse to staging. It allows for `INSERT IGNORE` to be executed.

warehouse/sql/V201702061486427077__initial_ddl.sql
agorina: Yes, but I think @malaffoon changed it because we had seen larger values.

warehouse/sql/V201702061486427077__initial_ddl.sql
```diff
@@ -240,45 +263,47 @@ CREATE TABLE IF NOT EXISTS item_difficulty_cuts (
...
+  first_name varchar(60),
   middle_name varchar(60),
-  gender_id tinyint NOT NULL,
```
chdietz: why did we take the not null off gender.

reporting/sql/V201702061486427077__initial_ddl.sql
agorina: Here is a pattern that I have decided to adopt: if a unique index included all columns in a table, then it will have just a table name. Otherwise I will list columns. Making changes to be consistent.

PR #98, opened 2017-07-05
DWR-464

malaffoon:
@agorina can you look at this please? I'm not sure why the `position` column types were set up this way, just want to verify with you that we can/should make this change.

### DIFF
--- a/reporting/sql/V201707051499276344__item_position.sql
+++ b/reporting/sql/V201707051499276344__item_position.sql
@@ -1,0 +1,7 @@
+# fix the item position column types
+
+USE ${schemaName};
+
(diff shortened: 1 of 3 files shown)


reporting/sql/V201707051499276344__item_position.sql
```diff
@@ -0,0 +1,7 @@
+# fix the item position column types
+
+USE ${schemaName};
+
+ALTER TABLE item MODIFY position smallint;
+
+ALTER TABLE exam_item MODIFY position smallint NOT NULL;
```
agorina: The reason I did not do it is because I do not think 'summative' will always have a position.

PR #101, opened 2017-07-07
DWR-287 Instructional resources storage

agorina:


reporting/sql/V201707071499465971__instructional_resource.sql
```diff
@@ -0,0 +1,11 @@
+# add instructional_resource
+
+USE ${schemaName};
+
+/**  Instructional resources to store mapping from assessment id to an external resource URL. **/
+/**  This table will be loaded/update manually. **/
+
+CREATE TABLE IF NOT EXISTS instructional_resource (
+  id int NOT NULL PRIMARY KEY,
+  resource varchar(1000) NOT NULL
```
malaffoon: Since it is a small table and we don't know much about the possible resource values, why not just make this `text`?

agorina: ```depth_of_knowledge``` table has a similar column with this size, so I copied it. But you bring up a good point, since we do not know the data - ```text``` is safer. Will change.

agorina:
@esotrope , yes -  join on resource.id=asmt.id.

PR #102, opened 2017-07-09
Bugfix/dwr 529

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_schema/102)


reporting/sql/V201707071499469492__namespace_constants.sql
```diff
@@ -0,0 +1,14 @@
+# add namespace column value constants for use in translation table.
+
+USE ${schemaName};
+
+
+# This creates a table that should have only 1 row.
+CREATE TABLE IF NOT EXISTS namespace_constants (
```
rcordeau: So the intent with this table is to have a table with only one row that has the namespace constant values so that I don't have to hard code the constant values in the Ingest .sql files. The need came about for wanting/having to delete "backend" rows only from reporting.translation when migrating from warehouse.

reporting/sql/V201707071499469491__translation_types.sql
```diff
@@ -0,0 +1,9 @@
+# modify translation table columns per DWR-529
+
+USE ${schemaName};
+
+#Increasing the size of the columns, but need to shrink the size of the prefix of lable_code that gets indexed
+ALTER TABLE translation MODIFY label text, 
+	MODIFY label_code varchar(256), 
+	DROP INDEX idx__translation__namespace_label_code_language_code,
+	ADD UNIQUE INDEX idx__translation__namespace_label_code_language_code (namespace, label_code(241), language_code);
```
rcordeau: varchar(256) was the size asked for on the label_code column. The problem with 256 is that it makes the unique index be larger than 767 bytes, so I kept the column size of 256 but shrunk the size of the prefix on that column that gets indexed.  I could always just shrink the size of the column instead.

agorina: I think we need label_code to be smaller in size. 
@bgsavage and @esotrope , could you please clarify why it has to be that long?

rcordeau: At first I had the label_code as large as it could be (varchar(241)), but then decided to shrink the prefix. I'm happy to make the column size smaller than 256.

reporting/sql/V201707071499469492__namespace_constants.sql
agorina: You lost me here. Could you please clarify why a one row table? 
Why not something like:
```CREATE TABLE IF NOT EXISTS translation_backend_namespace (
id tinyint NOT NULL PRIMARY KEY,
namespace varchar(10) NOT NULL );
And then every namespace in this table would be treated as being migrated from the warehouse?

Comment (rcordeau) [inline reporting/sql/V201707071499469492__namespace_constants.sql]: Yep, I could do that instead.  My first approach was just having a "contants" table for all namespaces. So, instead I could create the translation_backend_namespace table in the warehouse and modify the query to select from that table. Is that your preference?

Comment (agorina) [inline reporting/sql/V201707071499469492__namespace_constants.sql]: Yeah, except I think the `translation_backend_namespace` table belongs to the reporting or maybe both reporting and warehouse? The reporting needs to be aware of the namespaces that are reserved for the warehouse use.

File: reporting/sql/V201707071499469492__namespace_constants.sql [hunk 3]
```diff
@@ -0,0 +1,14 @@
+# add namespace column value constants for use in translation table.
```
malaffoon: Why separate sql files? Don't these changes go together?

reporting/sql/V201707071499469492__namespace_constants.sql
rcordeau: Sorry, I didn't mean warehouse, I meant reporting.

PR #107, opened 2017-07-18
add index on student ethnicity table

malaffoon:


warehouse/sql/V201707181500413929__student_ethnicity.sql
```diff
@@ -0,0 +1,6 @@
+-- add index to student_ethnicity to improve ingest performance of students
```
malaffoon: Per @agorina's suggestion, i made this a unique constraint on the join table. I verified that the query utilizes the index when querying against `student_id = :id` and improves performance significantly.

malaffoon: I also verified that none of the existing environments had data that would conflict with the unique constraint.

PR #108, opened 2017-07-19
hack for cleaning up names in ETS data

malaffoon:
More of an FYI than a review but ...

These are the scripts i used to clean the names in the ETS data (both UAT and QA).


warehouse/sql/clean_names_hack.sql
```diff
@@ -0,0 +1,146 @@
...
+  name      VARCHAR(60),
+  INDEX idx__last_name_mapping__hash_name(hash_name)
+);
+CREATE TABLE IF NOT EXISTS datagen.ssid_mapping (
+  hash_ssid VARCHAR(65),
+  ssid VARCHAR(10)
+);
+
+-- collect first "names" that we want to map
+INSERT INTO datagen.first_name_mapping (hash_name, gender_id)
```
malaffoon: Note that the primary mapping tables preserve both the old and new names so we can revert all this if we really need to.

warehouse/sql/clean_names_hack.sql
```diff
@@ -0,0 +1,146 @@
...
+UPDATE datagen.district_name_mapping SET name =
+  CONCAT_WS(' ',
+            (SELECT name FROM datagen.animal_names ORDER BY RAND() limit 1),
+            (SELECT name FROM datagen.animal_names ORDER BY RAND() limit 1),
+            'District')
+  WHERE name IS NULL;
+
+-- make an import record for these changes
+INSERT INTO warehouse.import (status, content, contentType, digest, batch)
+VALUES (0, 4, 'text/plan', 'MANUAL', 'mlaffoon cleaning organization data');
```
malaffoon: Thanks @agorina!

warehouse/sql/clean_names_hack.sql
agorina: Nice.

warehouse/sql/clean_names_hack.sql
agorina: :)

PR #111, opened 2017-07-25
add some missing common core standards

malaffoon:
There were some assessment items in the latest tabulator output that were failing due to missing common core standard entries in our lookup table. I used the spreadsheet @chdietz pointed me at to figure out all the missing math codes.

I didn't do english for two reasons: no failures during import and SB has chosen to arbitrarily swap hyphens and dots and i couldn't figure out the pattern. (They did the same thing with math codes but i could figure out the pattern.)


agorina:
I think @bgsavage will need to add them to the translation file.

PR #112, opened 2017-07-26
more common core codes

malaffoon:


warehouse/sql/V201707251501021717__more_code_cleanup.sql
```diff
@@ -0,0 +1,23 @@
+-- more code cleanup
+
+USE ${schemaName};
+
+-- Although we'd like to do this, we can't with existing data
+-- If a full data wipe is done (and scripts consolidated) most UNKNOWN claim targets should be eliminated
+-- DELETE FROM target WHERE description = 'UNKNOWN' AND natural_id NOT IN ('S-ID|P', 'G-SRT|O');
```
malaffoon: This is a little annoying. I think we're going to have to actually reload all assessments to get the corrected secondary claim targets. @agorina, perhaps we can discuss when we have some free time?

agorina: We can discuss it any time you want to.

warehouse/sql/V201707251501021717__more_code_cleanup.sql
```diff
@@ -0,0 +1,23 @@
...
+
+INSERT INTO common_core_standard (subject_id, natural_id, description) VALUES
```
chdietz: These are the code you discovered missing as mentioned I guess. Are any of these duplicates, in other words do they exist with another natural id?

malaffoon:
@chdietz, regarding the delete ...

When i import the latest tabulator output for the last three years, there are only 2 of the UNKNOWN claim targets referenced. However, in the existing databases there are 526 references to the UNKNOWN claim targets. It has to do with how the tabulator was parsing the assessment packages, that has subsequently been fixed.

warehouse/sql/V201707251501021717__more_code_cleanup.sql
malaffoon: Nope, they are all non-specific category-level codes that didn't exist. But, for example, G-CO.A was already in our table even though it is such a beast and really shouldn't be there. I guess AIR put it in.

PR #114, opened 2017-07-27
Instructional resources

agorina:
I am not sure what is the best way to handle it.
This is something that needs to be run AFTER asmts are loaded.

### DIFF
--- a/reporting/sql/Instructional_resources_dml.sql
+++ b/reporting/sql/Instructional_resources_dml.sql
@@ -1,0 +1,25 @@
+-- loads data AFTER asmts have been loaded into reporting
+-- original source is : http://www.smarterbalanced.org/educators/the-digital-library/
+
+INSERT IGNORE INTO reporting.instructional_resource (id, resource)
+SELECT id, 'https://portal.smarterbalanced.org/library/en/v1.0/digital-library-connections-grade-3-number-and-operations-in-base-ten.docx' AS resource FROM reporting.asmt WHERE label = 'Grade 03 Math - Number and Operations in Base Ten (IAB)'
+UNION ALL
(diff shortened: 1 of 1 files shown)


reporting/sql/Instructional_resources_dml.sql
```diff
@@ -0,0 +1,25 @@
+-- loads data AFTER asmts have been loaded into reporting
+-- original source is : http://www.smarterbalanced.org/educators/the-digital-library/
+
+INSERT IGNORE INTO reporting.instructional_resource (id, resource)
+SELECT id, 'https://portal.smarterbalanced.org/library/en/v1.0/digital-library-connections-grade-3-number-and-operations-in-base-ten.docx' AS resource FROM reporting.asmt WHERE label = 'Grade 03 Math - Number and Operations in Base Ten (IAB)'
```
agorina: Note that this returns multiple records (per grade).

malaffoon:
I'd approve this but i have no idea what/why/where (pretty clear on how/who though).

agorina:
@malaffoon , it is a phase 1 implementation of the Digital Library.
In short: we do not have requirements and SBAC does not really have a lot of resources built yet. As a shortcut, we are doing this. More can be found here: https://jira.fairwaytech.com/browse/DWR-287

reporting/sql/Instructional_resources_dml.sql
bgsavage: So `id` is `asmt.id` ?

agorina: That is correct.

bgsavage: Should we consider renaming this field to indicate so?

agorina: Good point. We should.

PR #116, opened 2017-07-27
DWR-637 Added index to import.

agorina:


warehouse/sql/V201707271501197885__import_created_updated_idx.sql
```diff
@@ -0,0 +1,5 @@
+-- migrate related changes : adding index to import
+
+USE ${schemaName};
+
+ALTER TABLE import ADD INDEX idx__import__updated_status (updated, status);
```
agorina: As part of changes to migrate I have this query: 

```
SELECT MAX(updated) as stop FROM (SELECT updated FROM import WHERE status != 0 AND (:prev IS NULL OR updated >= :prev) ORDER BY updated LIMIT :batchSize) as batch
```

PR #117, opened 2017-07-28
DWR-637 Added created to master entities

agorina:
Here is a use case why we need `created` on the "master" entities:

Create a new school at "2017-07-18 19:00:00"
Create a new exam for a student in the above school at ''2017-07-18 20:00:00"
Update the school from the above at "2017-08-18 19:00:00".

Try to re-migrate all of the above: unless there is a `created` timestamp, the migrate does not see the right sequence of events.

### DIFF
--- a/warehouse/sql/V201707271501213691__migrate_changes_add_created.sql
+++ b/warehouse/sql/V201707271501213691__migrate_changes_add_created.sql
@@ -1,0 +1,43 @@
+-- migrate related changes : adding created column to master entities
+
+USE ${schemaName};
+
+-- -------- asmt changes ------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


warehouse/sql/V201707271501213691__migrate_changes_add_created.sql
```diff
@@ -0,0 +1,43 @@
...
+UPDATE student s
+  JOIN import i ON i.id = s.import_id
+SET s.created = i.created;
+
+ALTER TABLE student DROP INDEX idx__student__updated;
+ALTER TABLE student ADD INDEX idx__student__created_uodated (created, updated);
+
+-- -------- student_group changes ------------------------------------------------------------------------------------------------
+ALTER TABLE student_group DROP INDEX idx__student_group__updated;
+ALTER TABLE student_group ADD INDEX idx__student_group__created_updated (created, updated);
```
agorina: `student_group` already has `created`.

PR #119, opened 2017-07-31
trick flyway

malaffoon:
I'm not sure how/when i will merge this, the intent is to test whether flyway can be tricked into believing that a consolidated SQL script is the current baseline for an existing database.

I took a fully populated database with the latest set of scripts applied. I ran the flway_hack script on it. And, indeed, flyway was happy and believed the state of the database was good and proper.

The problem, of course, is that the flyway_hack script must be run by hand during deployment to any  environment. So there is a bit of coordination. Still, i think it is worth it to produce "version 1.0.0" scripts when we release 1.0.0.


agorina:
I like the idea.

PR #121, opened 2017-07-31
DWR-637 Part 2 - changes to support a timestamp based migrate

agorina:


reporting/sql/V201707311501538157__migrate_changes_part2.sql
```diff
@@ -0,0 +1,28 @@
+-- migrate related changes : removing unused columns
+
+USE ${schemaName};
+
+-- NOTE: before applying this script all migrates have to be completed
```
agorina: This will only work if everything with the smaller timestamp have been migrated prior to the release.

reporting/sql/V201707311501538157__migrate_changes_part2.sql
```diff
@@ -0,0 +1,28 @@
...
+UPDATE migrate SET first_at = CURRENT_TIMESTAMP(6) WHERE first_at IS NULL;
+UPDATE migrate SET last_at = CURRENT_TIMESTAMP(6) WHERE last_at IS NULL;
+
+ALTER TABLE migrate MODIFY COLUMN first_at TIMESTAMP(6) NOT NULL;
+ALTER TABLE migrate MODIFY COLUMN last_at TIMESTAMP(6) NOT NULL;
+
+ALTER TABLE migrate DROP COLUMN first_import_id;
+ALTER TABLE migrate DROP COLUMN last_import_id;
+
+UPDATE asmt SET updated = CURRENT_TIMESTAMP(6) WHERE updated IS NULL;
```
malaffoon: Not sure it matters but this got me thinking: before we could associate a master object (asmt, exam, etc.) with a migrate record by using the import_id. That goes away. Just noticed because there is no definitive way to set `updated` based on the migrate record here. Again, it doesn't matter that much ...

reporting/sql/V201707311501538157__migrate_changes_part2.sql
malaffoon: I'm nervous about this being a Vxxx script if running it is conditional. We have to be careful is all.

reporting/sql/V201707311501538157__migrate_changes_part2.sql
agorina: This does not really go away. We kept the import_id in the master tables in the warehouse and it is migrated to the reporting too.

PR #122, opened 2017-08-01
DWR-637 Final (hopefully final) touch.

agorina:


reporting/sql/V201708011501603952__migrate_add_size_and_codes_flag.sql
```diff
@@ -0,0 +1,8 @@
+-- more migrate related changes
+
+USE ${schemaName};
+
+ALTER TABLE migrate ADD INDEX idx__migrate__status_last_at (status, last_at);
```
agorina: This seems to be the best option for SQL like this:
```SELECT MAX(m.last_at) FROM reporting.migrate m WHERE m.status = 20```

PR #123, opened 2017-08-02
DWR-694 Add report generation table

wtritch:
Add the report_generation table to the reporting schema.

I waffled a bit on the name: "report" sounded too generic inside a "reporting" schema.  Thoughts?

### DIFF
--- a/reporting/sql/V201708011501635215__add_report_generation.sql
+++ b/reporting/sql/V201708011501635215__add_report_generation.sql
@@ -2,15 +2,14 @@
 
 USE ${schemaName};
 
-CREATE TABLE IF NOT EXISTS report_generation (
+CREATE TABLE IF NOT EXISTS user_report (
   id bigint(20) NOT NULL AUTO_INCREMENT PRIMARY KEY,
-  user_login varchar(255) COLLATE utf8_unicode_ci NOT NULL,
+  user_login varchar(255) NOT NULL,
   job_execution_id bigint(20),
   status tinyint(4) NOT NULL,
   report_resource_uri varchar(255),
   label varchar(255) NOT NULL,
   report_request text NOT NULL,
   created TIMESTAMP(6) NOT NULL,
-  KEY idx__report_generation__user_login_label (user_login, label),
-  KEY idx__report_generation__created (created)
+  INDEX idx__report_generation__user_login (user_login)
 );
--- a/reporting/sql/V201708011501635215__add_report_generation.sql
+++ b/reporting/sql/V201708011501635215__add_report_generation.sql
@@ -1,0 +1,16 @@
+-- Add a report_generation table to track user report requests
+
+USE ${schemaName};
+
+CREATE TABLE IF NOT EXISTS report_generation (
+  id bigint(20) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  user_login varchar(255) COLLATE utf8_unicode_ci NOT NULL,
+  job_execution_id bigint(20),
+  status tinyint(4) NOT NULL,
+  report_resource_uri varchar(255),
+  label varchar(255) NOT NULL,
+  report_request text NOT NULL,
+  created TIMESTAMP(6) NOT NULL,
+  KEY idx__report_generation__user_login_label (user_login, label),
+  KEY idx__report_generation__created (created)
+);


reporting/sql/V201708011501635215__add_report_generation.sql
```diff
@@ -0,0 +1,16 @@
+-- Add a report_generation table to track user report requests
+
+USE ${schemaName};
```
wtritch: I'm guessing this will be used to also create the table in the test schema?

agorina: Yes, this is correct.

reporting/sql/V201708011501635215__add_report_generation.sql
```diff
@@ -0,0 +1,16 @@
...
+
+CREATE TABLE IF NOT EXISTS report_generation (
+  id bigint(20) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  user_login varchar(255) COLLATE utf8_unicode_ci NOT NULL,
```
agorina: You do not need COLLATE here, it is configured on the schema level.

reporting/sql/V201708011501635215__add_report_generation.sql
```diff
@@ -0,0 +1,16 @@
...
+  job_execution_id bigint(20),
+  status tinyint(4) NOT NULL,
+  report_resource_uri varchar(255),
+  label varchar(255) NOT NULL,
+  report_request text NOT NULL,
+  created TIMESTAMP(6) NOT NULL,
+  KEY idx__report_generation__user_login_label (user_login, label),
```
agorina: Would you want it to be UNIQUE?

reporting/sql/V201708011501635215__add_report_generation.sql
```diff
@@ -0,0 +1,16 @@
...
+  KEY idx__report_generation__created (created)
```
agorina: I am not exactly clear how your query looks, but it seems that ```created``` is part of the ORDER BY, while the other two columns are in the WHERE clause. Is this the case?  Or is it used independently?

jtreuting: I would assume that the ID could be used for sorting and created would just be for display purposes and so maybe the index on created is not needed?

reporting/sql/V201708011501635215__add_report_generation.sql
jtreuting: I'm guessing not unique since the label should be auto-generated (or at least auto-generated the suggestion) and it might be something like "School A Grade 3 Report" or maybe including the date.

malaffoon: use of `KEY` is inconsistent with other sql we have, please use `INDEX` unless there is a compelling reason

reporting/sql/V201708011501635215__add_report_generation.sql
```diff
@@ -0,0 +1,16 @@
+-- Add a report_generation table to track user report requests
+
+USE ${schemaName};
+
+CREATE TABLE IF NOT EXISTS report_generation (
```
malaffoon: For name, perhaps `user_report` or `report_request`. But this is fine too.

reporting/sql/V201708011501635215__add_report_generation.sql
wtritch: nice, removed.

reporting/sql/V201708011501635215__add_report_generation.sql
wtritch: I'd vote for not unique, since the label is user-enterable (with a suggested name).  I wanted label in the index in case we allow the user to search for their reports by label... but that may be premature optimization.  I'll remove the label from the index for now.

reporting/sql/V201708011501635215__add_report_generation.sql
wtritch: good point, I can order by auto-generated id rather than create date.  I removed the created index, and will change the SQL to order by id.

reporting/sql/V201708011501635215__add_report_generation.sql
wtritch: Changed to INDEX

reporting/sql/V201708011501635215__add_report_generation.sql
wtritch: I like user_report.  Changed.

PR #124, opened 2017-08-02
DWR-689 Merge reporting schemas

agorina:
I am abandoning all the batch data since there is no FK anywhere and I do not think it is useful to migrate them.  I am not sure how to automate the `drop` of the old schemas. I guess we can do it manually, unless you know a better way?

After looking into it more I am considering adding `migrate_id` to the master tables in the reporting. This column exists in the `staging` tables but I think it would be helpful (for troubleshooting) to have it in the final destinations. Please let me know what you think.


build.gradle
```diff
@@ -76,8 +76,6 @@ distributions {
             from ('.') {
                 include 'reporting/**'
                 include 'warehouse/**'
-                include 'staging/**'
-                include 'spring_batch_reporting/**'
                 include 'LICENSE'
                 include 'build.gradle'
                 include 'gradle.properties'
```
agorina: I am not sure if this is enough to clean up gradle?

malaffoon: Given the philosophy of this change, i suggest we get rid of "datamart" here: its only purpose was to group the three schemas.

malaffoon:
Since we're changing the Vxxx.sql files, i think perhaps i should do the merge and manual rebase of everything.

agorina:
@malaffoon, I would like to merge it first and then you can do your trick. Will this work for you?

build.gradle
```diff
@@ -62,10 +62,6 @@ def ts = project.getTasks()
             group = "$allTask$type"
             description = "Custom task to " + allTask.capitalize() + MessageFormat.format(" all objects in the configured {0} schemas.", type ? "test" : "production/development")
         }
-        task("${allTask}DataMart$type", dependsOn: [ts.findAll{x-> x.name in dataMartSchemas.collect {allTask + it.capitalize() + type}}]) {
-            group = "$allTask$type"
-            description = "Custom task to " + allTask.capitalize() +  MessageFormat.format(" all objects in the configured {0} data mart schemas.", type ? "test" : "production/development")
-        }
     }
 }
```
agorina: @malaffoon , does this look right?

malaffoon: seems ok

PR #126, opened 2017-08-03
DWR-545 Initial schema for student group upload

agorina:


warehouse/sql/V201708021501717811__upload_student_group.sql
```diff
@@ -0,0 +1,73 @@
...
+  name varchar(20) NOT NULL UNIQUE
+);
+
+INSERT INTO upload_student_group_status (id, name) VALUES
+  (-4, 'UNAUTHORIZED'),
+  (-3, 'BAD_DATA'),
+  (-2, 'BAD_FORMAT'),
+  (-1, 'INVALID'),
+  (0, 'ACCEPTED'),
+  (1, 'PROCESSED');
```
agorina: I copied these statuses from import. I am not sure is we need INVALID?

warehouse/sql/group_admin_hack.sql
```diff
@@ -2,8 +2,8 @@ USE warehouse;
 
 ########################################### schema changes ###########################################
 
-DROP TABLE IF EXISTS student_group_load;
-CREATE TABLE IF NOT EXISTS student_group_load (
+DROP TABLE IF EXISTS upload_student_group;
+CREATE TABLE IF NOT EXISTS upload_student_group (
```
agorina: I have tried to keep the names in synch, but this script is most likely broken now since some of the columns/constraints may not work.

@chdietz, please let me know if you would prefer to have a working version but with the different table names from the script above.

warehouse/sql/V201708021501717811__upload_student_group.sql
malaffoon: do we need a separate set of statuses?

warehouse/sql/V201708021501717811__upload_student_group.sql
```diff
@@ -0,0 +1,73 @@
...
+  
+CREATE TABLE IF NOT EXISTS upload_student_group_batch (
+  id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  digest varchar(32) NOT NULL,
+  status tinyint NOT NULL,
+  creator varchar(250),
+  created timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  updated timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+  message text,
+  INDEX idx__import__digest (digest)
```
malaffoon: copy and paste name error

warehouse/sql/V201708021501717811__upload_student_group.sql
agorina: Good point. We can share, removing this.

warehouse/sql/V201708021501717811__upload_student_group.sql
agorina: Fixing.

warehouse/sql/group_admin_hack.sql
chdietz: I think we(you) should use the table names we discussed with upoad_ prefix.
If something is broken due to changes in others tables i can resolve, i hope :)

warehouse/sql/V201708021501717811__upload_student_group.sql
chdietz: Mark, are you suggesting use the same status such as import_status ?

warehouse/sql/group_admin_hack.sql
agorina: Sorry, could you please clarify if you want me to keep the change in this file or leave it without changes? :)

chdietz: I would keep the changes and add a comment that it needs to be tested and there may be issues constraints.

agorina: Done, thank you!

PR #129, opened 2017-08-17
Record original filename when uploading a Group definition CSV.

wtritch:
There is a request from the UI team to display the original filename in the upload results table.  This PR adds a column to the upload_student_group_batch table representing the upload so that we have a place to store it.

### DIFF
--- a/warehouse/sql/V201708171503010368__add_student_group_batch_filename.sql
+++ b/warehouse/sql/V201708171503010368__add_student_group_batch_filename.sql
@@ -1,0 +1,5 @@
+-- Modify upload_student_group_batch to include uploaded file name
+
+USE ${schemaName};
+
+ALTER TABLE upload_student_group_batch ADD filename VARCHAR(200);
--- a/warehouse/sql/V201708171503010368__add_student_group_batch_filename.sql
+++ b/warehouse/sql/V201708171503010368__add_student_group_batch_filename.sql
@@ -2,4 +2,4 @@
 
 USE ${schemaName};
 
-ALTER TABLE upload_student_group_batch ADD filename VARCHAR(200);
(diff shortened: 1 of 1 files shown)


warehouse/sql/V201708171503010368__add_student_group_batch_filename.sql
```diff
@@ -0,0 +1,5 @@
+-- Modify upload_student_group_batch to include uploaded file name
+
+USE ${schemaName};
+
+ALTER TABLE upload_student_group_batch ADD filename VARCHAR(200);
```
wtritch: I'll just do something like:
```
StringUtils.abbreviateMiddle(filename, "...", 150);
```
in the associated model builder to keep the filename under 200 characters.

bgsavage: Isn't 255 the max file size on most operating systems anyway?

malaffoon: Hmm, i guess. Seems like we should decide: are we storing just the filename or do we want to try to preserve the entire pathname. If it's just the filename, then make it VARCHAR(250) and hope for the best. If it is the pathname, make it TEXT and forget about it.

wtritch: I don't think I want the pathname, even if I could get it.  I'll bump this to 255.  But I think my paranoia is still going to make me limit the string length in code =P

bgsavage: Yeah I'd still limit the length in code.  Also agree that we don't need the original path name.

PR #130, opened 2017-08-22
DWR-503 Legacy data load - first draft of the schema

agorina:
This is just the first draft. Wanted to save it so that @chdietz could use it for the data load.


build.gradle
```diff
@@ -72,6 +73,7 @@ distributions {
             from ('.') {
                 include 'reporting/**'
                 include 'warehouse/**'
+                include 'legacy_load/**'
                 include 'LICENSE'
                 include 'build.gradle'
                 include 'gradle.properties'
```
agorina: Am I missing anything here?

legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -0,0 +1,268 @@
...
+** 	Initial script for the SBAC Reporting Data Legacy Load schema
+**
+** Open Questions:
+** 1. How to related old asm to the loaded : asmt_guid?
+** 2. How to map the claims (JSON or asmt table)
+** 3. For the overlapping data (school id, student id: what to use -db native or natural)
+** 4. Student data: what takes precedence : dim_student or fact tables? (for those that overlap)
+** 5. Opportunity - derive from the same asmt diff date?
+** 6. session id - hardcode to 'legacy' or derive?
+**
```
agorina: I am planning on running some queries on the production to answer the above questions. (WIth the exception of the item 2, @chdietz  nicely agreed to research it).

legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -0,0 +1,268 @@
...
+CREATE TABLE IF NOT EXISTS dim_inst_hier (
+  inst_hier_rec_id bigint NOT NULL PRIMARY KEY,
+  state_code varchar(2) NOT NULL,
+  district_id varchar(40) NOT NULL,
+  district_name varchar(60) NOT NULL,
+  school_id varchar(40) NOT NULL,
+  school_name varchar(60) NOT NULL,
+  from_date varchar(8) NOT NULL,
+  to_date varchar(8),
+#   rec_status varchar(1) NOT NULL,
```
agorina: I did not want to remove the columns yet, just commented out the candidates to be deleted from this schema. I am assuming that we will load the active status only and hence this is not needed.

legacy_load/sql/V201708211503359336__initial_ddl.sql
chdietz: :)

build.gradle
```diff
@@ -34,11 +34,12 @@ def flywayActions = [migrate: org.flywaydb.gradle.task.FlywayMigrateTask,
 
 def reportingSchemas = ["reporting"]
 def warehouseSchemas = ["warehouse"]
+def legacyLoadSchemas = ["legacy_load"]
 
 //Gradle can't call a single task more than once, because of its design, so
 //we dynamically create individual tasks for each of the available flyway commands
 //on all of our databases (dev and test). Task names end up like migrateStaging and cleanWarehouse_test.
-(reportingSchemas + warehouseSchemas).each { schema ->
+(reportingSchemas + warehouseSchemas + legacyLoadSchemas).each { schema ->
```
malaffoon: uhmm, i wouldn't enable the gradle tasks at all for this; we don't want to be running these scripts in any automated fashion do we?

agorina:
It is convenient while I am developing to have one command to recreate a db. I agree that we will not run it this way in the production.

PR #132, opened 2017-08-22
DWR-503 Legacy data load - open task breakdown

agorina:
I am using this review to get feedback on the open questions and any comments/concerns about the closed items.

### DIFF
--- a/legacy_load/sql/V201708211503359336__initial_ddl.sql
+++ b/legacy_load/sql/V201708211503359336__initial_ddl.sql
@@ -14,7 +14,7 @@
 **   http://mysql.rjweb.org/doc.php/deletebig
 ** 8. how to assign import ids for migrate. Note: in the import table somehow capture that this is legacy migration
 ** 9. add support for the 'catch up' load
-** 10. Mapping of the old accommodations to the new codes:
+** 10. [DONE] Mapping of the old accommodations to the new codes:
 **  acc_asl_video_embed smallint NOT NULL,                          TDS_ASL1
 **  acc_braile_embed smallint NOT NULL,                             ENU-Braille
 **  acc_closed_captioning_embed smallint NOT NULL,                  TDS_ClosedCap1
--- a/legacy_load/sql/V201708211503359336__initial_ddl.sql
+++ b/legacy_load/sql/V201708211503359336__initial_ddl.sql
@@ -1,13 +1,35 @@
 /**
 ** 	Initial script for the SBAC Reporting Data Legacy Load schema
 **
-** Open Questions:
-** 1. How to related old asm to the loaded : asmt_guid?
+** Open Items:
+** 1. How to relate old asm to the loaded one: asmt_guid does not work.
 ** 2. How to map the claims (JSON or asmt table)
-** 3. For the overlapping data (school id, student id: what to use -db native or natural)
-** 4. Student data: what takes precedence : dim_student or fact tables? (for those that overlap)
-** 5. Opportunity - derive from the same asmt diff date?
-** 6. session id - hardcode to 'legacy' or derive?
+** 3. [DONE] For the overlapping data - does not matter, seem to be identical
+** 4. [DONE] Student data: what takes precedence : dim_student or fact tables? - does not matter, seem to be identical
+** 5. [DONE] Opportunity - derive from the same asmt, date_taken_year and student and diff date, start from 0 and increment
+** 6. [DONE] Session id - hardcode to 'legacy'
+** 7. what is the largest chunk of records in one transaction and how to chunk the data for updates/inserts?
+**   stored procedure or java?
+**   some resources to check out: https://stackoverflow.com/questions/29754570/bulk-update-to-an-unindexed-column-in-a-large-innodb-table
+**   http://mysql.rjweb.org/doc.php/deletebig
+** 8. how to assign import ids for migrate. Note: in the import table somehow capture that this is legacy migration
+** 9. add support for the 'catch up' load
+** 10. Mapping of the old accommodations to the new codes:
+**  acc_asl_video_embed smallint NOT NULL,                          TDS_ASL1
+**  acc_braile_embed smallint NOT NULL,                             ENU-Braille
+**  acc_closed_captioning_embed smallint NOT NULL,                  TDS_ClosedCap1
+**  acc_text_to_speech_embed smallint NOT NULL,                     TDS_TTS_Stim&amp;TDS_TTS_Item
+**  acc_abacus_nonembed smallint NOT NULL,                          NEA_Abacus
+**  acc_alternate_response_options_nonembed smallint NOT NULL,      NEA_AR
+**  acc_calculator_nonembed smallint NOT NULL,                      NEA_Calc
+**  acc_multiplication_table_nonembed smallint NOT NULL,            NEA_MT
+**  acc_print_on_demand_nonembed smallint NOT NULL,                 TDS_PoD_Stim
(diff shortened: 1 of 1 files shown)


legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -1,13 +1,35 @@
 /**
 ** 	Initial script for the SBAC Reporting Data Legacy Load schema
 **
-** Open Questions:
-** 1. How to related old asm to the loaded : asmt_guid?
+** Open Items:
+** 1. How to relate old asm to the loaded one: asmt_guid does not work.
```
agorina: This is an unknown. I do not see any data that could clearly relate the asmts. I hope I am just missing something obvious.

legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -1,13 +1,35 @@
...
 ** 2. How to map the claims (JSON or asmt table)
```
agorina: There are two options: one to hardcode the value based on the JSON template and another is to use the asmt table. I feel that the second is probably better, but would like to get the confirmation.

legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -1,13 +1,35 @@
...
-** 3. For the overlapping data (school id, student id: what to use -db native or natural)
-** 4. Student data: what takes precedence : dim_student or fact tables? (for those that overlap)
-** 5. Opportunity - derive from the same asmt diff date?
-** 6. session id - hardcode to 'legacy' or derive?
+** 3. [DONE] For the overlapping data - does not matter, seem to be identical
+** 4. [DONE] Student data: what takes precedence : dim_student or fact tables? - does not matter, seem to be identical
+** 5. [DONE] Opportunity - derive from the same asmt, date_taken_year and student and diff date, start from 0 and increment
```
agorina: I am concern that this may be slow, just a note.

legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -1,13 +1,35 @@
...
+** 6. [DONE] Session id - hardcode to 'legacy'
+** 7. what is the largest chunk of records in one transaction and how to chunk the data for updates/inserts?
+**   stored procedure or java?
+**   some resources to check out: https://stackoverflow.com/questions/29754570/bulk-update-to-an-unindexed-column-in-a-large-innodb-table
+**   http://mysql.rjweb.org/doc.php/deletebig
+** 8. how to assign import ids for migrate. Note: in the import table somehow capture that this is legacy migration
+** 9. add support for the 'catch up' load
+** 10. [DONE] Mapping of the old accommodations to the new codes:
```
agorina: @jtreuting , thank you for this mapping.

legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -1,13 +1,35 @@
...
-** 5. Opportunity - derive from the same asmt diff date?
-** 6. session id - hardcode to 'legacy' or derive?
+** 3. [DONE] For the overlapping data - does not matter, seem to be identical
+** 4. [DONE] Student data: what takes precedence : dim_student or fact tables? - does not matter, seem to be identical
+** 5. [DONE] Opportunity - derive from the same asmt, date_taken_year and student and diff date, start from 0 and increment
+** 6. [DONE] Session id - hardcode to 'legacy'
+** 7. what is the largest chunk of records in one transaction and how to chunk the data for updates/inserts?
+**   stored procedure or java?
+**   some resources to check out: https://stackoverflow.com/questions/29754570/bulk-update-to-an-unindexed-column-in-a-large-innodb-table
+**   http://mysql.rjweb.org/doc.php/deletebig
```
agorina: If anybody has an answer to this or any additional resource, pease share.

legacy_load/sql/V201708211503359336__initial_ddl.sql
malaffoon: we can do a screen share tomorrow and look around; that would be scary and disappointing

legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -1,13 +1,35 @@
...
+** 1. How to relate old asm to the loaded one: asmt_guid does not work.
 ** 2. How to map the claims (JSON or asmt table)
-** 3. For the overlapping data (school id, student id: what to use -db native or natural)
-** 4. Student data: what takes precedence : dim_student or fact tables? (for those that overlap)
-** 5. Opportunity - derive from the same asmt diff date?
-** 6. session id - hardcode to 'legacy' or derive?
+** 3. [DONE] For the overlapping data - does not matter, seem to be identical
+** 4. [DONE] Student data: what takes precedence : dim_student or fact tables? - does not matter, seem to be identical
+** 5. [DONE] Opportunity - derive from the same asmt, date_taken_year and student and diff date, start from 0 and increment
+** 6. [DONE] Session id - hardcode to 'legacy'
```
jtreuting: If this is going to be the value that is displayed to the user in the reporting application, I think it should be "unknown" instead.

legacy_load/sql/V201708211503359336__initial_ddl.sql
```diff
@@ -1,13 +1,35 @@
...
+** 7. what is the largest chunk of records in one transaction and how to chunk the data for updates/inserts?
+**   stored procedure or java?
+**   some resources to check out: https://stackoverflow.com/questions/29754570/bulk-update-to-an-unindexed-column-in-a-large-innodb-table
+**   http://mysql.rjweb.org/doc.php/deletebig
+** 8. how to assign import ids for migrate. Note: in the import table somehow capture that this is legacy migration
```
jtreuting: Might not be possible, but my first thought is to use negative import IDs because they would be obvious and not get in the way of others moving forward.

legacy_load/sql/V201708211503359336__initial_ddl.sql
agorina: Let's do that. I am hoping that Venue may have these data if we do not find it.

legacy_load/sql/V201708211503359336__initial_ddl.sql
agorina: "unknown" works.

legacy_load/sql/V201708211503359336__initial_ddl.sql
agorina: I do not feel comfortable with the negative ids. It will break our integration tests. I was thinking about using some of the text fields (such as batch or digest) in the import table to capture this info.

legacy_load/sql/V201708211503359336__initial_ddl.sql
agorina: I would like to chat about it. 

Since we do not know the session, all the opportunities will be in one session 'unknown'. Does it make sense?

legacy_load/sql/V201708211503359336__initial_ddl.sql
chdietz: I thought negative import id's used to work in integration tests for migrate, did that change?

agorina: No, it did not change. The negative ids are used in the tests and that is the reason why I would not want to use them in the real life.

PR #134, opened 2017-08-22
extract and old scripts for one table to test end to end flow

chdietz:
starting with dim_inst_hier to test end to end flow.
already have extracted and in S3
tested load with local non S3 load
will look into bash script for running and copy to s3 and tracking multiple runs in S3

### DIFF
--- a/legacy_load/sql/extract_dim_inst_hier.sql
+++ b/legacy_load/sql/extract_dim_inst_hier.sql
@@ -1,0 +1,12 @@
+-- example run
+-- psql -h localhost -U edware -d edware -a -f extract_dim_inst_hier.sql
+
+-- use the ca schema in srl-db-master-003
+set SEARCH_PATH to ca;
+
+-- use the edware schema to test in dev
+-- SET SEARCH_PATH TO edware;
+
+-- Extract current records to CSV
+-- copy does not like multiline statements
+\copy (SELECT i.inst_hier_rec_id, i.state_code, i.district_id, i.district_name, i.school_id, i.school_name, i.from_date, i.to_date, i.rec_status, i.batch_guid FROM dim_inst_hier i WHERE i.rec_status = 'C') TO '/mnt/pgsql/rdw-migrate/extract_dim_inst_hier.csv' WITH CSV HEADER;
--- a/legacy_load/sql/load_dim_inst_hier.sql
+++ b/legacy_load/sql/load_dim_inst_hier.sql
@@ -1,0 +1,17 @@
+-- Load extracted records from CSV
+
+use legacy_load;
+
+-- Test loading local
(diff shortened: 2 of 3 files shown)


chdietz:
should have been extract and load scripts not old scripts

legacy_load/sql/extract_dim_inst_hier.sql
```diff
@@ -0,0 +1,12 @@
...
+
+-- use the ca schema in srl-db-master-003
+set SEARCH_PATH to ca;
+
+-- use the edware schema to test in dev
+-- SET SEARCH_PATH TO edware;
+
+-- Extract current records to CSV
+-- copy does not like multiline statements
+\copy (SELECT i.inst_hier_rec_id, i.state_code, i.district_id, i.district_name, i.school_id, i.school_name, i.from_date, i.to_date, i.rec_status, i.batch_guid FROM dim_inst_hier i WHERE i.rec_status = 'C') TO '/mnt/pgsql/rdw-migrate/extract_dim_inst_hier.csv' WITH CSV HEADER;
```
malaffoon: Should we just be getting all columns?  It would make your single line statement shorter  ;)

legacy_load/sql/load_dim_inst_hier.sql
```diff
@@ -0,0 +1,17 @@
...
+-- INTO TABLE dim_inst_hier
+-- FIELDS TERMINATED BY ',' IGNORE 1 LINES
+-- (inst_hier_rec_id, state_code, district_id, district_name, school_id, school_name, from_date, to_date, rec_status, batch_guid)
+-- SET warehouse_load_id = 1;
+
+-- Load from S3
+LOAD DATA FROM S3 's3://rdw-migrate/extract_dim_inst_hier.csv'
+INTO TABLE dim_inst_hier
+FIELDS TERMINATED BY ',' IGNORE 1 LINES
+(inst_hier_rec_id, state_code, district_id, district_name, school_id, school_name, from_date, to_date, rec_status, batch_guid)
```
malaffoon: again, can we just do all columns? Obviously we'd have to make sure the table schema lines up.

agorina: We need to check how boolean is looking. I did talk with Chris offline. In postgres I am seeing t/f and I need them to be 1/0 in mySQL. This may require the casting.

chdietz: yes, I will check boolean as I hit them

legacy_load/sql/extract_dim_inst_hier.sql
chdietz: ah, yes, good point we went back and forth on that.
we landed on all columns but explicitly calling them out in case the order was changed in the new schema

PR #140, opened 2017-08-23
Update the upload_student_group name column to match required CSV header

wtritch:
I decided it was easier to change the DB to match RDW_Admin vs the reverse.

### DIFF
--- a/warehouse/sql/V201708221503439275__update_group_upload_schema.sql
+++ b/warehouse/sql/V201708221503439275__update_group_upload_schema.sql
@@ -1,0 +1,6 @@
+-- Modify upload_student_group "group_name" column to match csv headers
+
+USE ${schemaName};
+
+ALTER TABLE upload_student_group
+  CHANGE name group_name VARCHAR(255);


warehouse/sql/V201708221503439275__update_group_upload_schema.sql
```diff
@@ -0,0 +1,6 @@
+-- Modify upload_student_group "group_name" column to match csv headers
+
+USE ${schemaName};
+
+ALTER TABLE upload_student_group
+  CHANGE name group_name VARCHAR(255);
```
malaffoon: One has to wonder at the fascination with 2^n-1 values in today's modern world.  ;)

chdietz: hmm I think Alla had a reason for this I asked about it

wtritch: It's just as arbitrary as base-10. =)  If we had 6 fingers on each hand, we'd be counting in twelves =P

PR #143, opened 2017-08-29
DWR-758 Datamart: add school_year table to datamart

agorina:
I still need to update migrate and reporting to use the tables.

Sorry for the messy review, please ignore the first commit. It is a left over from the legacy load that I forgot to push.

### DIFF
--- a/legacy_load/sql/post-validation.sql
+++ b/legacy_load/sql/post-validation.sql
@@ -114,15 +114,15 @@ INSERT INTO post_validation(testNum, result1, result2, result3, result4)
 INSERT INTO post_validation(testNum, result1, result2) SELECT (SELECT max(testNum) FROM post_validation) + 1, 'ethnicity count', 'ethnicity';
 INSERT INTO post_validation(testNum, result1, result2)
   SELECT (SELECT max(testNum) FROM post_validation),
-    count(*), ethnicity_code  FROM student_ethnicityGROUP BY ethnicity_codeORDER BY count(*);
+    count(*), ethnicity_code  FROM student_ethnicity GROUP BY ethnicity_code ORDER BY count(*);
 
 -- Exam accommodations
 INSERT INTO post_validation(testNum, result1, result2) SELECT (SELECT max(testNum) FROM post_validation) + 1, 'ica accommodations count', 'ethnicity';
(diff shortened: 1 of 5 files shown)


legacy_load/sql/post-validation.sql
```diff
@@ -114,15 +114,15 @@ INSERT INTO post_validation(testNum, result1, result2, result3, result4)
 INSERT INTO post_validation(testNum, result1, result2) SELECT (SELECT max(testNum) FROM post_validation) + 1, 'ethnicity count', 'ethnicity';
```
agorina: unrelated, forgot to push this change yesterday

legacy_load/sql/pre-validation.sql
```diff
@@ -15,15 +15,6 @@ CREATE TABLE IF NOT EXISTS pre_validation
   created timestamp default current_timestamp
 );
 
--- TODO: this is an asmt that was not loaded into prod yet
```
agorina: ditto

reporting/sql/V201708291504028600__add_school_year_table.sql
```diff
@@ -0,0 +1,19 @@
...
+  year smallint NOT NULL PRIMARY KEY
+);
+
+-- This should not be part of the initial db load, instead it should be treated as 'CODES' and migrated.
+-- But since we have the data loaded, in order to define constraints, this is loaded.
+INSERT INTO school_year (year) VALUES
+  (2015),
+  (2016),
+  (2017),
+  (2018);
```
agorina: From production:
```
mysql> select distinct school_year from exam;
+-------------+
| school_year |
+-------------+
|        2016 |
|        2017 |
|        2015 |
+-------------+
```

esotrope: Just curious, why are the values hardcoded and not based on the select above?

agorina: Ideally it should be, BUT it gets a little complicated and slow. 

School year is in three tables, so if we want it to be dynamic it should come from all of them. 

The data flow in continuously and the new school_year could get loaded in any time, so we would have to do it with every single migrate. 

To keep it simple I chose the route of preloading the value ( I believe this is what the JIRA stated).

PR #146, opened 2017-08-31
consolidate scripts for 1.0.0 release

malaffoon:
When i check this in i will have to hand-jam the flyway schema table in each environment. Developers will have to wipe and reload or do the same flyway trick. I will generate a mysql dump image using these scripts as well.


chdietz:
With a wipe and reload, or gradle clean and migrate, there is no need for any flyway tricks or hand-jam's?

malaffoon:
@chdietz that is correct.

PR #147, opened 2017-08-31
DWR-953 Performance: add completed_at to the index since it is used in ORDER BY

agorina:


reporting/sql/V1_0_0_2__add_completed_at_to_exam_index.sql
```diff
@@ -0,0 +1,10 @@
+-- add completed_at to the index since it is used in the ORDER BY
+
+USE ${schemaName};
+
+ALTER TABLE reporting.exam
```
malaffoon: schema prefix shouldn't be there ... won't work as expected for reporting_test

agorina: Thank you, 'copy and paste' error.

PR #148, opened 2017-09-01
DWR-953 Added index to import to support code migration

agorina:


warehouse/sql/V1_0_0_2__add_index_to_import.sql
```diff
@@ -0,0 +1,5 @@
+-- add index to support code migration
+
+USE ${schemaName};
+
+ALTER TABLE import ADD INDEX idx__import__content_status_created(content, status, created);
```
agorina: While analyzing the slow query logs I ran across this one:
```
SELECT 1 AS existsCodeImportContent
FROM warehouse.import wi WHERE wi.content = 3
  AND ((wi.status != 0 AND wi.updated >= '2017-08-30 19:31:41.345884' AND wi.updated <= '2017-08-31 04:00:05.333946') OR (wi.status = 0 AND wi.created <= '2017-08-31 04:00:05.333946')) LIMIT 1;
```
This query is executed with every migrate to determine if CODEs need to be migrates, so it is better to be fast. 

Trying to run it on the dev takes a long time ( I did not wait long enough to get the exact number).  This changes it to 233ms.

But I am concern about adding an extra index to import. 
There are currently three indexes defined:
```  
  INDEX idx__import__digest (digest),
  INDEX idx__import__created (created),
  INDEX idx__import__updated_status (updated, status)
```
I am tempted to remove *idx__import__created* since I do not think it is used, but then it is so close to the release that I am not comfortable with making this change without retesting.

What do you think?

malaffoon: I'd leave it in. We can remove it and test next week with all our free time.

PR #151, opened 2017-09-01
insert or update the html.system-news content that is displayed on th…

jtreuting:
…e homepage


RunBook.md
```diff
@@ -77,7 +77,7 @@ mysql> SELECT LAST_INSERT_ID() into @IMPORT_ID;
 - To complete the process, change the status of the import to 1:
 ```sql
 # trigger migration
-mysql> UDATE import SET status = 1 WHERE id = @IMPORT_ID;
+mysql> UPDATE import SET status = 1 WHERE id = @IMPORT_ID;
```
jtreuting: unrelated, just happened to notice it

reporting/sql/V201709011504300509__update_homepage_systemnews.sql
```diff
@@ -0,0 +1,12 @@
+# insert or update the html.system-news content for the homepage
+
+USE ${schemaName};
+
+REPLACE INTO translation (label_code, namespace, language_code, label) VALUES ('html.system-news', 'frontend', 'eng', '<h2 class="blue-dark h3 mb-md">Note</h2>
```
jtreuting: used replace into because i wasn't sure if there would be a value there already or not

reporting/sql/V201709011504300509__update_homepage_systemnews.sql
```diff
@@ -0,0 +1,12 @@
+# insert or update the html.system-news content for the homepage
```
malaffoon: need to use the new numbering; depending on which of us checks in first, this should be V1_0_0_3__upda... or V1_0_0_4__upda...

jtreuting: just saw this. i made my addition before pulling the latest so all the files had the old way still.  i'll change to V1_0_0_3__ ... cool?

malaffoon: yup, i'll wait for your changes

reporting/sql/V201709011504300509__update_homepage_systemnews.sql
malaffoon: good choice

reporting/sql/V1_0_0_3__update_homepage_systemnews.sql
```diff
@@ -0,0 +1,12 @@
+# insert or update the html.system-news content for the homepage
```
jtreuting: @malaffoon just noticed that the new scripts have `-- comment` instead of the pound sign.  should i change this?

malaffoon: meh, no need

RunBook.md
agorina: Thank you!

PR #152, opened 2017-09-01
patch unknown targets

malaffoon:
Turns out that the UNKNOWN targets were duplicates of targets with invalid ids. So, the descriptions for the UNKNOWN targets are updated and the targets with invalid ids are removed. I verified that there are no items referencing the targets with invalid ids.

### DIFF
--- a/reporting/sql/V1_0_0_4__fix_targets.sql
+++ b/reporting/sql/V1_0_0_4__fix_targets.sql
@@ -1,0 +1,14 @@
+-- fixes two errant targets
+--
+-- instead of relying on migrate, this just patches in the changes
+-- (when consolidating scripts that means you can just ignore this file)
+
+USE ${schemaName};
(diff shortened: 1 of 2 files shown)


reporting/sql/V1_0_0_4__fix_targets.sql
```diff
@@ -0,0 +1,14 @@
+-- fixes two errant targets
+--
+-- instead of relying on migrate, this just patches in the changes
+-- (when consolidating scripts that means you can just ignore this file)
+
```
agorina: Don't you need to update the initial dml in order to ignore this file during scripts consolidation?

malaffoon: yeah, but that'll be taken care of when the warehouse scripts is consolidated

PR #155, opened 2017-09-05
DWR-961 Tune up SQL for monitoring system health

agorina:


Monitoring.md
```diff
@@ -34,10 +36,8 @@ Each ingest is different and hence the processing time will vary, but in general
 
 To monitor for slow imports:
 ```sql
-SELECT count(*)
- FROM import
- WHERE status = 0 AND TIMESTAMPDIFF(SECOND, CURRENT_TIMESTAMP(6), updated) > 60;
```
agorina: Using functions does not work well with the indexes. 
The recommended way is to do it like this, so I changed it everywhere here.

PR #159, opened 2017-09-18
Add chunk processing tracking to the user_report table.

wtritch:
Now that we're doing cross-pod fan-out during report generation, we need a shared point for tracking how many student "chunks" have been processed when generating a large report.  This PR adds "total" and "completed" columns to the user_report table.  It also removes the job_execution_id column that will no longer be used once the report processor is converted from spring batch to spring stream processing.

### DIFF
--- a/reporting/sql/V1_1_0_1__add_user_report_chunk_tracking.sql
+++ b/reporting/sql/V1_1_0_1__add_user_report_chunk_tracking.sql
@@ -1,6 +1,5 @@
--- Modify user_report table to track chunk completion and remove unused job_execution_id column
+-- Modify user_report table to track chunk completion
 USE ${schemaName};
 
-ALTER TABLE user_report DROP COLUMN job_execution_id;
 ALTER TABLE user_report ADD COLUMN total_chunk_count INT NOT NULL DEFAULT 0;
 ALTER TABLE user_report ADD COLUMN complete_chunk_count INT NOT NULL DEFAULT 0;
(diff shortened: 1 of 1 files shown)


reporting/sql/V1_1_0_1__add_user_report_chunk_tracking.sql
```diff
@@ -0,0 +1,6 @@
+-- Modify user_report table to track chunk completion and remove unused job_execution_id column
+USE ${schemaName};
+
+ALTER TABLE user_report DROP COLUMN job_execution_id;
```
wtritch: The more I think about it, the more dangerous this seems.  Once I remove this column we cannot push a schema change until after the report-processor changes are deployed.  Should I remove this from here and add a cleanup PR once the report-processor instances have been updated?

wtritch: I'm going to leave the job_execution_id column in for now and will file a PR to remove it once the changes are fully deployed.

agorina:
I do not see anything in regards to " It also removes the job_execution_id column that will no longer be used once the report processor is converted from spring batch to spring stream processing." ?

wtritch:
@agorina I decided to remove that piece since it prevents us from pushing a schema change until after all the report-processor changes are ready for deployment.

agorina:
Makes sense, thank you for the clarifications.

PR #161, opened 2017-09-20
add groups of districts/schools

malaffoon:


build.gradle
```diff
@@ -18,7 +18,7 @@ apply plugin: 'distribution'
 apply plugin: 'maven-publish'
 
 group = 'org.opentestsystem.rdw.schema'
-version = '1.0.0-SNAPSHOT'
+version = '1.1.0-SNAPSHOT'
```
malaffoon: think it is time to start bumping release versions in the develop branches

warehouse/sql/V1_1_0_1__add_school_groups.sql
```diff
@@ -0,0 +1,35 @@
+-- add district and school groups
```
malaffoon: district, district_group and school_group are all flat tables with just the metadata for the organization. No attempt is being made to preserve the relationship between them. Only the school table records information about ancestry.

warehouse/sql/V1_1_0_1__add_school_groups.sql
```diff
@@ -0,0 +1,35 @@
...
+  ADD COLUMN school_group_id int,
+  ADD COLUMN external_id varchar(40),
+  ADD INDEX idx__school__district_group (district_group_id),
+  ADD INDEX idx__school__school_group (school_group_id),
+  ADD CONSTRAINT fk__school__district_group FOREIGN KEY (district_group_id) REFERENCES district_group (id),
+  ADD CONSTRAINT fk__school__school_group FOREIGN KEY (school_group_id) REFERENCES school_group (id);
+
+-- although these procedures aren't used by the code any more they are used in some legacy_load scripts
+-- DROP PROCEDURE IF EXISTS school_upsert;
+-- DROP PROCEDURE IF EXISTS district_upsert;
```
agorina: I would move the code to the lgeacy_load schema/scripts and drop them from the warehouse schema.

warehouse/sql/V1_1_0_1__add_school_groups.sql
```diff
@@ -0,0 +1,35 @@
...
+CREATE TABLE IF NOT EXISTS school_group (
+  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  natural_id varchar(40) NOT NULL,
+  name varchar(100) NOT NULL,
+  external_id varchar(40),
+  UNIQUE INDEX idx__school_group__natural_id (natural_id)
+);
+
+ALTER TABLE district
+  ADD COLUMN external_id varchar(40);
```
agorina: Could you please explain what this is? 
Is this the id SBAC wanted for the mapping to the other external data?

warehouse/sql/V1_1_0_1__add_school_groups.sql
jtreuting: i don't see a relationship between a district and district_group.  I'm thinking of the District Level Export which is more than that really and needs to allow for the ability to handle someone that has a Group of District role.  I think they would need to see what districts they have access to so they could export all results for a district.  as i'm typing, i guess there is the school to district relationship which could be used to determine this.

malaffoon: That is what i was thinking: to do a group-of-districts export you find all test results for schools that have that group-of-district in their ancestry.

warehouse/sql/V1_1_0_1__add_school_groups.sql
malaffoon: Yes, this is an arbitrary external id that will be populated with the NCES id from ART.

warehouse/sql/V1_1_0_1__add_school_groups.sql
malaffoon: Ok, i'll push that around and resubmit.

warehouse/sql/V1_1_0_1__add_school_groups.sql
```diff
@@ -0,0 +1,35 @@
...
+ALTER TABLE school
+  ADD COLUMN district_group_id int,
+  ADD COLUMN school_group_id int,
+  ADD COLUMN external_id varchar(40),
+  ADD INDEX idx__school__district_group (district_group_id),
+  ADD INDEX idx__school__school_group (school_group_id),
+  ADD CONSTRAINT fk__school__district_group FOREIGN KEY (district_group_id) REFERENCES district_group (id),
+  ADD CONSTRAINT fk__school__school_group FOREIGN KEY (school_group_id) REFERENCES school_group (id);
+
+-- code no longer uses these; this change is not backward compatible!
```
malaffoon: I'll have to coordinate pushing code changes with this schema change. So i'll probably sit on this PR until the code is complete and reviewed.

reporting/sql/V1_1_0_2__add_school_groups.sql
```diff
@@ -0,0 +1,56 @@
+-- add district and school groups
```
malaffoon: @agorina i added this, can you take a quick look?

reporting/sql/V1_1_0_2__add_school_groups.sql
```diff
@@ -0,0 +1,56 @@
...
+
+USE ${schemaName};
+
+CREATE TABLE IF NOT EXISTS staging_district_group (
+  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
agorina: AUTO_INCREMENET should not be in the staging tables.

PR #164, opened 2017-09-28
Feature/add unique exam index

wtritch:
This PR:
1. de-dupes exam key values by modifying the existing oppId
2. creates a unique oppId for exams with a NULL oppId

For de-duping we should only be modifying hand-jammed data in our test environments.
For creating unique oppId values we will be populating legacy production exams where the oppId was not persisted in the legacy system that provided our original exam data.

Migration on my local machine with ~11.5k exams took ~3.6 seconds (for this migration as well as the previous migration of merging of exam_student and exam.)
Extrapolating out linearly that gives us roughly ~2.5hrs to process the 25 million exam records in awsdev.

### DIFF
--- a/warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
+++ b/warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
@@ -68,5 +68,6 @@ ALTER TABLE exam
   DROP COLUMN partition_id,
   DROP INDEX idx__exam__asmt_id_student_id_oppId;
 
+-- add unique index on exam key properties
 ALTER TABLE exam
   ADD UNIQUE INDEX idx__exam__asmt_id_student_id_oppId (asmt_id, student_id, oppId);
--- a/warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
+++ b/warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
@@ -34,7 +34,7 @@ DELIMITER ;
 -- disable updates to the 'updated' and add a partition and index for processing
 ALTER TABLE exam
   MODIFY COLUMN updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
-  ADD INDEX idx__exam__asmt_id_student_id_oppId (asmt_id, student_id, oppId),
+  ADD INDEX idx__exam__asmt_student_oppId (asmt_id, student_id, oppId),
   ADD COLUMN partition_id int;
 
 SET @exam_student_partition_start = 0;
@@ -66,8 +66,8 @@ DROP PROCEDURE loop_by_partition;
 ALTER TABLE exam
   MODIFY COLUMN updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
   DROP COLUMN partition_id,
-  DROP INDEX idx__exam__asmt_id_student_id_oppId;
+  DROP INDEX idx__exam__asmt_student_oppId;
 
 -- add unique index on exam key properties
 ALTER TABLE exam
-  ADD UNIQUE INDEX idx__exam__asmt_id_student_id_oppId (asmt_id, student_id, oppId);
+  ADD UNIQUE INDEX idx__exam__asmt_student_oppId (asmt_id, student_id, oppId);
--- a/warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
+++ b/warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
@@ -1,0 +1,72 @@
+-- Correct exam.oppId conflicts:
+-- 1: Generate oppIds for legacy data with NULL values
+-- 2: Dissociate exams with the same asmt_id, oppId, student_id by modifying
+--    their existing oppId values
+
+USE ${schemaName};
+
+-- Helper to run partitions in a loop
+DELIMITER //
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
```diff
@@ -0,0 +1,73 @@
...
+      SET iteration = iteration + 1;
+      IF iteration <= p_last THEN
+        ITERATE partition_loop;
+      END IF;
+      LEAVE partition_loop;
+    END LOOP partition_loop;
+
+  END;
+//
+DELIMITER ;
```
wtritch: This technique is a duplicate of @agorina's implementation from the previous migration =)

warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
```diff
@@ -0,0 +1,73 @@
...
+
+-- disable updates to the 'updated' and add a partition and index for processing
+ALTER TABLE exam
+  MODIFY COLUMN updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  ADD INDEX idx__exam__asmt_id_student_id_oppId (asmt_id, student_id, oppId),
+  ADD COLUMN partition_id int;
```
wtritch: The temporary index significantly speeds up processing.

warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
```diff
@@ -0,0 +1,73 @@
...
+    END LOOP partition_loop;
+
+  END;
+//
+DELIMITER ;
+
+-- disable updates to the 'updated' and add a partition and index for processing
+ALTER TABLE exam
+  MODIFY COLUMN updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  ADD INDEX idx__exam__asmt_id_student_id_oppId (asmt_id, student_id, oppId),
```
agorina: It is not going to change anything but the approach on the index names being implemented now is that if it is a FK to a PK on another table, do not use 'id' in the name. In other words this index would be named: "idx__exam__asmt_student_oppId".

wtritch: works for me, changed

agorina: You may want to move 'oppId' to the first position in the index. 
It may help with the second call when you have only one  'oppId' in the where clause.

wtritch: I'll give it a shot and see if performance improves

warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
```diff
@@ -0,0 +1,73 @@
...
+
+-- enable auto updates to the 'updated'
+ALTER TABLE exam
+  MODIFY COLUMN updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+  DROP COLUMN partition_id,
+  DROP INDEX idx__exam__asmt_id_student_id_oppId;
+
+-- add unique index on exam key properties
+ALTER TABLE exam
+  ADD UNIQUE INDEX idx__exam__asmt_id_student_id_oppId (asmt_id, student_id, oppId);
```
agorina: I would recommend changing the order here to by moving student_id to the first position.
May help with finding this record for the update faster.

warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
wtritch: Slight improvement from 3.6s down to 3.4s =)

warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
wtritch: Will do, but I think we already have a FK index on student_id

warehouse/sql/V1_1_0_3__correct_oppid_conflicts.sql
```diff
@@ -0,0 +1,73 @@
...
+
+-- Dissociate duplicates
+CALL loop_by_partition(
+    'UPDATE exam e1
+      JOIN exam e2 ON e2.id != e1.id
+        AND e2.asmt_id = e1.asmt_id
+        AND e2.student_id = e1.student_id
+        AND e2.oppId = e1.oppId
+    SET e1.oppId = CONCAT(CAST(e1.id as CHAR), ''_'', e1.oppId)
+    WHERE e1.oppId IS NOT NULL', @exam_student_partition_start, @exam_student_partition_end);
```
malaffoon: How does this change data when there are no dups, one dup, two dups, etc.? I assume if there are no dups, there is no change. If there are dups are all the records updated?

wtritch: Yessir, if there are no dups then the data is unchanged.
The de-duping is performed on a partitioned basis.  So, if the last duplicate value is in a separate partition it will be unchanged since there are no dups by the time the migration reaches it.

So original duplicate values like:
```
id, oppId
1, "myId"
27, "myId"
496, "myId"
```
could become
```
1, "1_myId"
27, "myId"
496, "496_myId"
```
with the final oppId only being guaranteed to be unique.

malaffoon: coolio, thanks for clarifying

PR #166, opened 2017-10-03
DWR-389 - Triggers and audit tables for exam and exam child tables

chdietz:
Starting with exam and related child tables. 
The following audit tables are created:
audit_exam
audit_exam_claim_score
audit_exam_available_accommodation
audit_exam_item

Each table has:
  id - the primary key of the table
  action - 'delete' or 'update'
  audited - timestamp when audit record created
  database_user - the MySql user name and host ( so we could filter DBA vs Application )
  Then each field in the source table.

The tables being audited have triggers, currently:
exam - update (delete is soft)
exam_claim_score - update
exam_available_accommodation - delete
exam_item - update and delete

If we stick with triggers I think it may consider having update and delete triggers on all audited
tables.  For now I am only adding triggers for what our application does to the tables.

There are sample trt's that exercise triggers on the confluence design page.
Can get to it from the Jira also.
There are sequence numbers on trt's so they submit in order of hypothetical changes.

One concern I have is no trigger 'mute' feature for when updates may occur during a flyway migration.  Options others have mentioned include.
1. drop triggers and re add.  if we did this we would want a separate trigger drop and create script
2. set a variable to disable triggers and check if it is on in the trigger sql.

### DIFF
--- a/warehouse/sql/V1_1_0_4__exam_audit.sql
+++ b/warehouse/sql/V1_1_0_4__exam_audit.sql
@@ -9,6 +9,9 @@ USE ${schemaName};
   Audit records are created for update and delete.
   Timestamp is required for exam updates that include creating a new child record.
   Child records do now have the import id.
+
+  Through the application, after insert, some exam records can be updated, some deleted or both as noted below.
+  Triggers are added consistently for update and delete on all exam tables.
 */
 
 ALTER TABLE exam_claim_score
@@ -63,6 +66,7 @@ CREATE TABLE IF NOT EXISTS audit_exam (
   prim_disability_type VARCHAR(3) NULL
 );
 
+-- UPDATE
 CREATE TRIGGER trg__exam__update
 BEFORE UPDATE ON exam
 FOR EACH ROW
@@ -108,6 +112,51 @@ FOR EACH ROW
       OLD.language_code,
       OLD.prim_disability_type);
 
+-- DELETE
+CREATE TRIGGER trg__exam__delete
+BEFORE DELETE ON exam
+FOR EACH ROW
+  INSERT INTO audit_exam (action, database_user, exam_id, type_id, school_year, asmt_id, asmt_version,
+                          opportunity, oppId, completeness_id, administration_condition_id, session_id, scale_score,
+                          scale_score_std_err, performance_level, completed_at, import_id, update_import_id, deleted,
+                          created, updated, grade_id, student_id, school_id, iep, lep, section504,
+                          economic_disadvantage, migrant_status, eng_prof_lvl, t3_program_type, language_code,
+                          prim_disability_type)
+  VALUES
+    (
+      'delete',
+      USER(),
+      OLD.id,
+      OLD.type_id,
+      OLD.school_year,
+      OLD.asmt_id,
+      OLD.asmt_version,
+      OLD.opportunity,
+      OLD.oppId,
+      OLD.completeness_id,
+      OLD.administration_condition_id,
+      OLD.session_id,
+      OLD.scale_score,
+      OLD.scale_score_std_err,
+      OLD.performance_level,
+      OLD.completed_at,
+      OLD.import_id,
+      OLD.update_import_id,
+      OLD.deleted,
+      OLD.created,
+      OLD.updated,
+      OLD.grade_id,
+      OLD.student_id,
+      OLD.school_id,
+      OLD.iep,
+      OLD.lep,
+      OLD.section504,
+      OLD.economic_disadvantage,
+      OLD.migrant_status,
+      OLD.eng_prof_lvl,
+      OLD.t3_program_type,
+      OLD.language_code,
+      OLD.prim_disability_type);
 
 /*
   exam_claim_score audit table and triggers
@@ -127,6 +176,7 @@ CREATE TABLE IF NOT EXISTS audit_exam_claim_score (
   category TINYINT NULL
 );
 
+-- UPDATE
 CREATE TRIGGER trg__exam_claim_score__update
 BEFORE UPDATE ON exam_claim_score
 FOR EACH ROW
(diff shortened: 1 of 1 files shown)


jtreuting:
I've done option 2 in the past.  Basically have a settings type table that the insert into can join on in the trigger SQL.  i feel that is cleaner than option 1 if we need to suspend triggers during migrations or for other reasons.

chdietz:
I think I prefer settings table or a variable vs drop and re add also.
I am looking to see if there is an agreement that we need a way to disable triggers during migrations.

warehouse/sql/V1_1_0_4__exam_audit.sql
```diff
@@ -0,0 +1,255 @@
+-- Create tables to hold audit for update and delete of exams
+-- 1: create audit tables
+-- 2: create triggers
+
+#USE ${schemaName};
```
malaffoon: uncomment before commit?

warehouse/sql/V1_1_0_4__exam_audit.sql
```diff
@@ -0,0 +1,255 @@
...
+  section504 TINYINT NULL,
+  economic_disadvantage TINYINT NOT NULL,
+  migrant_status TINYINT NULL,
+  eng_prof_lvl VARCHAR(20) NULL,
+  t3_program_type VARCHAR(20) NULL,
+  language_code VARCHAR(3) NULL,
+  prim_disability_type VARCHAR(3) NULL
+);
+
+DROP TRIGGER IF EXISTS trg__exam__update;
```
malaffoon: nit (because i know we have other scripts that are like this): i prefer that flyway scripts make assumptions about things. This is the init script for this so those triggers *do not exist*. If the script executes and fails because the script already exists, then we want it to fail because something is wrong.

warehouse/sql/V1_1_0_4__exam_audit.sql
```diff
@@ -0,0 +1,255 @@
...
+  INSERT INTO audit_exam (action, audited, database_user, exam_id, type_id, school_year, asmt_id, asmt_version,
+                          opportunity, oppId, completeness_id, administration_condition_id, session_id, scale_score,
+                          scale_score_std_err, performance_level, completed_at, import_id, update_import_id, deleted,
+                          created, updated, grade_id, student_id, school_id, iep, lep, section504,
+                          economic_disadvantage, migrant_status, eng_prof_lvl, t3_program_type, language_code,
+                          prim_disability_type)
+  VALUES
+    (
+      'update',
+      NOW(),
```
malaffoon: You have the column defined to default to current timestamp, so isn't setting it to NOW() kind of redundant?

agorina:
I like the idea of the settings type table to disable the triggers.

warehouse/sql/V1_1_0_4__exam_audit.sql
chdietz: good catch, wonder how that worked locally because I tested with gradle

warehouse/sql/V1_1_0_4__exam_audit.sql
chdietz: ok, there is no create trigger if not exits, so I did this to follow the create table if not exists as close as possible.
you bring up a good point, i know this table should not exist and the trigger should not exist so in both cases why check?

warehouse/sql/V1_1_0_4__exam_audit.sql
chdietz: good point if the default is current timestamp I can leave it out of the insert

wtritch:
Hmm... do we *also* need a trigger on INSERT on the child tables?  Because an UPDATE to an Exam may just involve an INSERT on a child table?

malaffoon:
Regarding @wtritch's comment ...

How rigorous do we need to be for this? The requirement says something about identifying who did what to the data. To Will's example, we could not definitively say that Bob added the child record, only that, when Bob made the change, there were child records associated with the exam.

agorina:
To answer @malaffoon's question, I believe we have discussed that we would indirectly relate children entities changes based on the master's `import record created/updated` and the` audit record/created`. Something like this: this master record has been changed by Bob and these are the children changes that happened during this time. Would like to get @jtreuting confirmation on this.

malaffoon:
But what timestamp is used to determine that a child record was created at a certain time?

agorina:
Good point. 
I looks like we either need to have a timestamp on the children or audit children insert.

chdietz:
To insert audit on child vs adding a timestamp on child records:
The thing I do not like about audit insert is it breaks our pattern on not auditing inserts.
And it is a lot of data.

warehouse/sql/V1_1_0_4__exam_audit.sql
```diff
@@ -0,0 +1,342 @@
...
+
+ALTER TABLE exam_claim_score
+  ADD COLUMN created timestamp(6) default CURRENT_TIMESTAMP(6) not null;
+
+ALTER TABLE exam_available_accommodation
+  ADD COLUMN created timestamp(6) default CURRENT_TIMESTAMP(6) not null;
+
+ALTER TABLE exam_item
+  ADD COLUMN created timestamp(6) default CURRENT_TIMESTAMP(6) not null;
+
```
chdietz: Added created timestamp to child tables of exam to associate inserted child records with the responsible import on an update to the exam.

warehouse/sql/V1_1_0_4__exam_audit.sql
```diff
@@ -0,0 +1,342 @@
...
+      OLD.lep,
+      OLD.section504,
+      OLD.economic_disadvantage,
+      OLD.migrant_status,
+      OLD.eng_prof_lvl,
+      OLD.t3_program_type,
+      OLD.language_code,
+      OLD.prim_disability_type);
+
+-- DELETE
```
chdietz: consistent delete and update trigger on each exam table.

agorina:
Did you run any performance benchmarks to see how this change affects the ingest performance?

chdietz:
@agorina no, I did not, we should discuss further, I think we need to benchmark with update and then with update and audit triggers on.

agorina:
Since update is going to be infrequent events I am not concerned about it a lot. But yes, let's talk about it.

PR #168, opened 2017-10-12
Feature/audit samples

chdietz:
Reading for review.

Todo:
* list of student related tables
* student table example queries

### DIFF
--- a/warehouse/AuditQueries.md
+++ b/warehouse/AuditQueries.md
@@ -1,11 +1,38 @@
 # Auditing
-This document describes auditing in RDW and provides samples queries for analysing audit data.
+This document describes auditing in RDW and provides samples queries for analysing audit data
+
+## Intended Audience
+The intended audience should be familiar with database technology and querying a database with SQL.
+- **System and Database Administrators**: This document provides administrators information on what is audited in the warehouse and where it is stored.
+- **Developers and Analysts**: Developers and analysts with knowledge of SQL and the appropriate permissions can use this document as a guide to querying exam and student modifications.
+
+## Terminology
+- **Test Result**: When a student takes a test the results are transmitted to the data warehouse.  This submission is a test result.  It is for one instance of a student taking a given test.
+- **TRT**: Is an acronym for an instance of a test result in the Smarter Balanced [Test Results Transmission Format](http://www.smarterapp.org/specs/TestResultsTransmissionFormat.html) where the content adheres to the [Test Results Data Model](http://www.smarterapp.org/news/2015/08/26/DataModelAndSamples.html)
+- **Exam**: Each test result submitted or migrated from legacy data is stored as an exam in the data warehouse.
+- **warehouse schema**: The warehouse schema is the source of truth for reporting in the data warehouse and is used to populate user reporting and analytical reporting schemas. All schemas are defined in the [SmarterApp/RDW_Schema](https://github.com/SmarterApp/RDW_Schema) repository.  The warehouse schema is in the [SmarterApp/RDW_Schema/warehouse](https://github.com/SmarterApp/RDW_Schema/tree/develop/warehouse) folder.
+- **Entity State**: Auditing tracks changes.
+  - **Create**: A new entity such as an exam is added to the warehouse.  This is not audited, however, there is an import record that records attributes of the submission.
+  - **Update**: A request to change a previously created entity.  This is audited as an update.
+  - **Delete**: An entity is removed from the warehouse.  Does not occur for entities being audited such as exam.  Does occur for entity attributes stored in supporting child tables and is audited as a delete.
+  - **Soft Delete**: A request to delete an entity from the warehouse that is being audited is updated with it's deleted flag set to true.  This is audited as an update.
+- **Import**:  All inflows of data to the warehouse create an import record that stores attributes of the inflow including a timestamp and the authorized user.
+  
 
 ## What is audited?
-... add details ... look in confluence functional page and mention that student and
+The warehouse audits entity state changes for exams and student information.
+Warehouse Exam Tables:
+
+| Table                        | Description                                 | Entity Type | States                      |
+|------------------------------|---------------------------------------------|-------------|-----------------------------|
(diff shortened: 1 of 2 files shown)


warehouse/AuditQueries.md
```diff
@@ -0,0 +1,324 @@
+# Auditing
```
malaffoon: nit: i think `Audit.md` would be a better name, this is discussing the overall audit system

warehouse/AuditQueries.md
```diff
@@ -0,0 +1,324 @@
...
+This document describes auditing in RDW and provides sample queries for analysing audit data.
+
+## Intended Audience
```
malaffoon: nit: can you bump the header levels up one to be consistent with some of the other markdown files? In other words, the first title would be `## Auditing`, and this one would be `### Intended Audience`, etc. Apart from consistency, this allows us to have a main entry document that then links to the other documents as chapters.

warehouse/AuditQueries.md
```diff
@@ -0,0 +1,324 @@
...
+- **Modification**: update, delete and soft delete state changes
+- **Import**:  All inflows of data to the warehouse create an import record that stores attributes of the inflow including a timestamp and the authorized user.
+  
+
+## What is audited?
+The warehouse audits entity state changes for exams and student information.
+
+Warehouse Exam Tables:
+
+| Table                        | Description                                 | Entity Type | States                      |
```
malaffoon: What is `States` in this table?

warehouse/AuditQueries.md
```diff
@@ -0,0 +1,324 @@
...
+|------------------------------|---------------------------------------------|-------------|-----------------------------|
+| exam                         | One record per test result                  | Parent      | Create, Update, Soft Delete |
+| exam_available_accommodation | One record per exam available accommodation | Child       | Create, Delete              |
+| exam_claim_score             | One record per exam claim                   | Child       | Create, Update              |
+| exam_item                    | One record per exam item                    | Child       | Create, Update, Delete      |
```
malaffoon: This will be updated to include student tables?

warehouse/AuditQueries.md
```diff
@@ -0,0 +1,324 @@
...
+
+```mysql
+SELECT
+  s.ssid,
+  CONCAT(s.last_or_surname, ', ', first_name) student,
+  e.id exam_id,
+  e.oppId,
+  e.opportunity,
+  asmt.name assessment_name,
+  SUM(CASE WHEN ae.exam_id IS NOT NULL THEN 1 ELSE 0 END) exam_update_count,
```
malaffoon: Doesn't `WHERE ae.exam_id IS NOT NULL` make the case statement unnecessary?

warehouse/AuditQueries.md
```diff
@@ -0,0 +1,324 @@
...
+
+**Finding modifications to a students exams**
+The following query outputs one row for each modified exam for one student.  It includes the count of modifications and the date of the last change.
+
+The where clause can be changed to include multiple students.
+
+```mysql
+SELECT
+  s.ssid,
+  CONCAT(s.last_or_surname, ', ', first_name) student,
```
malaffoon: I wonder if we shouldn't encourage them to extract PII information. Could we leave the name out of the example? They can always figure out how to put it in there if they want it.

warehouse/AuditQueries.md
```diff
@@ -0,0 +1,324 @@
...
+  e.opportunity,
+  asmt.name assessment_name,
+  SUM(CASE WHEN ae.exam_id IS NOT NULL THEN 1 ELSE 0 END) exam_update_count,
+  MAX(ae.audited) last_update
+FROM exam e
+LEFT JOIN audit_exam ae ON e.id = ae.exam_id
+JOIN student s ON e.student_id = s.id
+JOIN asmt ON e.asmt_id = asmt.id
+WHERE ae.exam_id IS NOT NULL
+  AND ae.audited > '2017-10-11 09:45'AND ae.audited < '2017-10-12 09:45'
```
malaffoon: nit: whitespace before `AND`

minor nit: there is a `BETWEEN` in mysql

warehouse/AuditQueries.md
```diff
@@ -0,0 +1,324 @@
...
+         lep,
+         section504,
+         economic_disadvantage,
+         migrant_status,
+         eng_prof_lvl,
+         t3_program_type,
+         language_code,
+         prim_disability_type
+       FROM exam e
+       WHERE import_id != update_import_id
```
malaffoon: So this query will return no results if an exam has never been updated? Should we mention that?

warehouse/AuditQueries.md
chdietz: I agree.

warehouse/AuditQueries.md
chdietz: Sounds like a good idea.

warehouse/AuditQueries.md
chdietz: ah, I changed it to Ingest State Changes and added a terminology for ingest.
In this column I am describing what ingest 'does' to the tables.
We do audit a delete of exam for example, but ingest will not do that, it will 'soft delete' it, when we implement it.

warehouse/AuditQueries.md
chdietz: right I had two versions of this query and the case is not needed in this one, you are correct.
The other example has all for a student and a zero count indicates no updates.
Kind of relates to your other question below, i may just add another example.

warehouse/AuditQueries.md
chdietz: interesting thought, but even the SSID is PII.

warehouse/AuditQueries.md
chdietz: Changed to use between

warehouse/AuditQueries.md
chdietz: Added another example that helps make this clear.

PR #169, opened 2017-10-12
Update the exam table for client-approved uniqueness based upon asses…

wtritch:
…sment id and opportunity id.

Add a status_date column to the exam table to resolve same-time updates/inserts.

### DIFF
--- a/warehouse/sql/V1_1_0_5__record_exam_status_date.sql
+++ b/warehouse/sql/V1_1_0_5__record_exam_status_date.sql
@@ -6,21 +6,6 @@ USE ${schemaName};
 ALTER TABLE exam
   ADD COLUMN status_date TIMESTAMP(6) DEFAULT NULL,
   DROP INDEX idx__exam__student_asmt_oppId,
-  ADD INDEX idx__exam__oppId_asmt (oppId, asmt_id);
-
--- Disambiguate oppIds to allow for a unique (oppId, asmt_id) index
-UPDATE exam e1
-  JOIN exam e2
-    ON e1.id != e2.id
-    AND e1.asmt_id = e2.asmt_id
-    AND e1.oppId = e2.oppId
-  SET
-    e1.oppId = CONCAT(CAST(e1.id as CHAR), '_', e1.oppId);
-
-ALTER TABLE exam
-  DROP INDEX idx__exam__oppId_asmt;
-
-ALTER TABLE exam
   ADD UNIQUE INDEX idx__exam__oppId_asmt (oppId, asmt_id);
 
 -- Update audit table and triggers
--- a/warehouse/sql/V1_1_0_5__record_exam_status_date.sql
+++ b/warehouse/sql/V1_1_0_5__record_exam_status_date.sql
@@ -6,4 +6,106 @@ USE ${schemaName};
 ALTER TABLE exam
   ADD COLUMN status_date TIMESTAMP(6) DEFAULT NULL,
   DROP INDEX idx__exam__student_asmt_oppId,
-  ADD UNIQUE INDEX idx__exam__asmt_oppId (asmt_id, oppId);
+  ADD UNIQUE INDEX idx__exam__oppId_asmt (oppId, asmt_id);
+
+-- Update audit table and triggers
+ALTER TABLE audit_exam
+  ADD COLUMN status_date TIMESTAMP(6) DEFAULT NULL;
+
+-- UPDATE
+DROP TRIGGER trg__exam__update;
+
+CREATE TRIGGER trg__exam__update
+BEFORE UPDATE ON exam
+FOR EACH ROW
+  INSERT INTO audit_exam (action, database_user, exam_id, type_id, school_year, asmt_id, asmt_version,
+                          opportunity, oppId, completeness_id, administration_condition_id, session_id, scale_score,
+                          scale_score_std_err, performance_level, completed_at, import_id, update_import_id, deleted,
+                          created, updated, grade_id, student_id, school_id, iep, lep, section504,
+                          economic_disadvantage, migrant_status, eng_prof_lvl, t3_program_type, language_code,
+                          prim_disability_type, status_date)
+  VALUES
+    (
+      'update',
+      USER(),
+      OLD.id,
+      OLD.type_id,
+      OLD.school_year,
+      OLD.asmt_id,
+      OLD.asmt_version,
+      OLD.opportunity,
+      OLD.oppId,
+      OLD.completeness_id,
+      OLD.administration_condition_id,
+      OLD.session_id,
+      OLD.scale_score,
+      OLD.scale_score_std_err,
+      OLD.performance_level,
+      OLD.completed_at,
+      OLD.import_id,
+      OLD.update_import_id,
+      OLD.deleted,
+      OLD.created,
+      OLD.updated,
+      OLD.grade_id,
+      OLD.student_id,
+      OLD.school_id,
+      OLD.iep,
+      OLD.lep,
+      OLD.section504,
+      OLD.economic_disadvantage,
+      OLD.migrant_status,
+      OLD.eng_prof_lvl,
+      OLD.t3_program_type,
+      OLD.language_code,
+      OLD.prim_disability_type,
+      OLD.status_date);
+
+-- DELETE
+DROP TRIGGER trg__exam__delete;
+
+CREATE TRIGGER trg__exam__delete
+BEFORE DELETE ON exam
+FOR EACH ROW
+  INSERT INTO audit_exam (action, database_user, exam_id, type_id, school_year, asmt_id, asmt_version,
+                          opportunity, oppId, completeness_id, administration_condition_id, session_id, scale_score,
+                          scale_score_std_err, performance_level, completed_at, import_id, update_import_id, deleted,
+                          created, updated, grade_id, student_id, school_id, iep, lep, section504,
+                          economic_disadvantage, migrant_status, eng_prof_lvl, t3_program_type, language_code,
+                          prim_disability_type, status_date)
+  VALUES
+    (
+      'delete',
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_1_0_5__record_exam_status_date.sql
```diff
@@ -0,0 +1,9 @@
+-- Update exam table to record Opportunity statusDate attribute
+-- Modify unique index to only include assessment id and opportunity id
+
+USE ${schemaName};
+
+ALTER TABLE exam
+  ADD COLUMN status_date TIMESTAMP(6) DEFAULT NULL,
```
wtritch: @chdietz should I add this column to your audit tables/triggers?

agorina: When we first set up the schema I have tried to adhere to the conventions defined here:https://confluence.fairwaytech.com/display/SBAC1719/Team+Development+Conventions
One of them says to not use a word 'date' in the date data types: '-Naming should be <type>_at; For example, created_at not create_date." I am not sure I completely agree with that approach, but just wanted to point it out.

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
```diff
@@ -0,0 +1,9 @@
...
+  DROP INDEX idx__exam__student_asmt_oppId,
+  ADD UNIQUE INDEX idx__exam__asmt_oppId (asmt_id, oppId);
```
agorina: If you are going to use this index in a look-up I suggest to flip the order since oppId is more discrete.

wtritch: Does type matter?  oppId is a varchar and asmt_id is a long

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
wtritch: hm... "status_at" sounds strange... I could be convinced if you feel strongly?

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
agorina: I think it is specific to each case but a high cardinality indexes usually perform faster.

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
agorina: I have no strong feelings about it, fine either way.

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
wtritch: oky doke, I'll swap it around

agorina: The best is probably to test it on the aws dev and see which one looks better.

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
```diff
@@ -0,0 +1,111 @@
...
+USE ${schemaName};
+
+ALTER TABLE exam
+  ADD COLUMN status_date TIMESTAMP(6) DEFAULT NULL,
+  DROP INDEX idx__exam__student_asmt_oppId,
+  ADD UNIQUE INDEX idx__exam__oppId_asmt (oppId, asmt_id);
+
+-- Update audit table and triggers
+ALTER TABLE audit_exam
+  ADD COLUMN status_date TIMESTAMP(6) DEFAULT NULL;
```
wtritch: I updated the audit table and triggers to account for the additional column.

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
chdietz: @wtritch yes on audit tables and triggers, sorry just saw this now.
But I am not sure I understand this date being added.
Did we find a new race condition?

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
chdietz: looks good

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
```diff
@@ -0,0 +1,111 @@
+-- Update exam table to record Opportunity statusDate attribute
+-- Modify unique index to only include assessment id and opportunity id
+
+USE ${schemaName};
+
+ALTER TABLE exam
+  ADD COLUMN status_date TIMESTAMP(6) DEFAULT NULL,
+  DROP INDEX idx__exam__student_asmt_oppId,
+  ADD UNIQUE INDEX idx__exam__oppId_asmt (oppId, asmt_id);
```
chdietz: Does this index make sense?
The TRDD states: "A unique identifier of the student-test opportunity. (Typically a GUID)"
There is nothing about what asmt it is for.

I think if the oppId is the same as an existing exam this is an update correct?

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
wtritch: @chdietz It's the same race condition that was identified earlier with the solution that we discussed.  The problem is that processing TRTs is asynchronous and multi-threaded.  So if two TRTs come in referring to the same exam at roughly the same time, (insert and delete, insert and update, update and update, etc)  we need a way to identify which one represents the actual desired state.  
This PR picks the TRT with the latest Opportunity.statusDate value, with the idea that the Opportunity.statusDate is updated whenever a TRT is generated/changed.

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
wtritch: I wasn't involved in the deep discussions with the stakeholders, but I believe the client specifically said that the "unique exam identifier" should be (asmt_id + oppId).  @malaffoon @agorina ?

agorina: Yes, this is how I remember it. 
@chdietz, do you have a different recollection?

chdietz: @wtritch ok thanks

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
```diff
@@ -0,0 +1,126 @@
...
+  ADD INDEX idx__exam__oppId_asmt (oppId, asmt_id);
+
+-- Disambiguate oppIds to allow for a unique (oppId, asmt_id) index
+UPDATE exam e1
+  JOIN exam e2
+    ON e1.id != e2.id
+    AND e1.asmt_id = e2.asmt_id
+    AND e1.oppId = e2.oppId
+  SET
+    e1.oppId = CONCAT(CAST(e1.id as CHAR), '_', e1.oppId);
```
wtritch: While working on speed tests, I found duplicate (asmt_id, oppId) values in our generated awsdev data.  I'm now disambiguating values by:
1. Add a non-unique index on (oppId, asmt_id)
2. Update any exam that has at least one duplicate (oppId, asmt_id) by prepending the exam.id to the oppId
3. Remove the non-unique index on (oppId, asmt_id) ad replace it with a *unique* index on (oppId, asmt_id)

Thoughts?

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
```diff
@@ -6,6 +6,21 @@ USE ${schemaName};
...
```
agorina: Are we planning on running this in production?

malaffoon: We shouldn't have to.

wtritch: But it shouldn't hurt anything either.

agorina: Also, we now have auditing triggers. I think we need to disable them before we run it? Not sure if that feature is done, we were talking about a control table to turn the triggers off. Tagging @chdietz for this question.

chdietz: @agorina  Yes I think I added another story for the disable trigger, and we did talk about it and the consensus was a settings table. This is not implemented yet.

agorina: In this case the triggers need to be turned off, right?

It feels to me that maybe we should have a separate script file (excluded from the production release) to clean up the made-up test data.

wtritch: We're already deleting and re-creating the triggers in this script, we could just delete, process, re-create.
As for creating a separate generated-only script file, we had the same debate w.r.t. V1_1_0_3__correct_oppid_conflicts.sql and decided to keep it all in the flyway scripts.  Has our reasoning changed?

agorina: I guess the real question is: if at the time of the production release we find duplicates, do we want to run this script (and if this is the case, then we should include it) or we want to stop the release?

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
wtritch: I reverted the disambiguation portion of the script in favor of running it manually against our generated-data environments: dev, qa, uat.

warehouse/sql/V1_1_0_5__record_exam_status_date.sql
wtritch: After discussion, I ran the disambiguation against our generated-data environments: dev, qa, uat and removed the functionality from this script.

PR #170, opened 2017-10-13
Feature/audit trigger switch dwr 1087

chdietz:
1. add a table to store the switch
2. modify the triggers to respect the switch

### DIFF
--- a/warehouse/sql/V1_1_0_6__exam_audit.sql
+++ b/warehouse/sql/V1_1_0_6__exam_audit.sql
@@ -37,8 +37,7 @@ FOR EACH ROW
                           created, updated, grade_id, student_id, school_id, iep, lep, section504,
                           economic_disadvantage, migrant_status, eng_prof_lvl, t3_program_type, language_code,
                           prim_disability_type, status_date)
-  VALUES
-    (
+  SELECT
       'update',
       USER(),
       OLD.id,
@@ -73,7 +72,8 @@ FOR EACH ROW
       OLD.language_code,
       OLD.prim_disability_type,
       OLD.status_date
-    );
+    FROM audit_trigger_switch ats
+    WHERE ats.id = 1 AND ats.switch = 1;
 
 -- DELETE
 DROP TRIGGER trg__exam__delete;
@@ -87,8 +87,7 @@ FOR EACH ROW
                           created, updated, grade_id, student_id, school_id, iep, lep, section504,
                           economic_disadvantage, migrant_status, eng_prof_lvl, t3_program_type, language_code,
                           prim_disability_type, status_date)
-  VALUES
-    (
+  SELECT
       'delete',
       USER(),
       OLD.id,
@@ -123,7 +122,8 @@ FOR EACH ROW
       OLD.language_code,
       OLD.prim_disability_type,
       OLD.status_date
-    );
+    FROM audit_trigger_switch ats
+    WHERE ats.id = 1 AND ats.switch = 1;
 
 /*
   exam_claim_score audit triggers
@@ -138,8 +138,7 @@ FOR EACH ROW
   INSERT INTO audit_exam_claim_score (
     action, database_user, exam_claim_score_id, exam_id, subject_claim_score_id, scale_score, scale_score_std_err, category
   )
-  VALUES
-    (
+  SELECT
       'update',
       USER(),
       OLD.id,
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_1_0_6__exam_audit.sql
```diff
@@ -0,0 +1,279 @@
...
+
+/*
+ Create the table to hold the on/off switch
+*/
+CREATE TABLE audit_trigger_switch (
+  id TINYINT NOT NULL PRIMARY KEY,
+  switch TINYINT NOT NULL,
+  updated timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6)
+);
+
```
chdietz: Not sure what people think about the name?
I put in a timestamp so you can see last time someone changed it.

warehouse/sql/V1_1_0_6__exam_audit.sql
```diff
@@ -0,0 +1,279 @@
...
+      OLD.economic_disadvantage,
+      OLD.migrant_status,
+      OLD.eng_prof_lvl,
+      OLD.t3_program_type,
+      OLD.language_code,
+      OLD.prim_disability_type,
+      OLD.status_date
+    FROM audit_trigger_switch ats
+    WHERE ats.id = 1 AND ats.switch = 1;
+
```
chdietz: This is where it looks for the trigger switch.  Same switch is used for all audit triggers.

warehouse/sql/V1_1_0_6__exam_audit.sql
```diff
@@ -0,0 +1,279 @@
...
+  id TINYINT NOT NULL PRIMARY KEY,
+  switch TINYINT NOT NULL,
+  updated timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6)
+);
+
+/*
+ Insert the single switch record in the 'on' position.
+ Disable audit triggers with: UPDATE audit_trigger_switch ats SET ats.switch = 0 WHERE ats.id = 1;
+ Enable audit triggers with:  UPDATE audit_trigger_switch ats SET ats.switch = 1 WHERE ats.id = 1;
+*/
```
chdietz: Tested changing this with a running ingest and works just fine.

warehouse/sql/V1_1_0_6__exam_audit.sql
agorina: Why do we need an id here? (just curious)

chdietz: do you mean why have two fields? id and switch?
i wanted it as simple as possible but I preferred to have an id that does not change
and another column for the value of the switch.

malaffoon: Maybe we should just have a generic `settings` table with key/value columns?

chdietz: @malaffoon I actually considered this and went back and forth.  So I am open to this.

agorina: I like this idea.

chdietz: The argument I thought of for the `audit_trigger_switch` table is:
* we have no immediate need for other settings and we should not build generic functionality we do not need.
* this setting makes up for the fact there is no `disable/enabale triggers` in MySQL
* when looking at the schema this table appears alphabetically near the audit tables.

The argument for generic settings table is:
* it provides a place for other settings if we need them
* the name of the setting is in the key, this may be more readable

If we go with a generic settings table then what do you think for the key and value?
* key = 'AUDIT_TRIGGER_ENABLE'
* value = 'TRUE' and triggers will insert audit records.
* value = anything other than 'TRUE' and triggers will NOT insert audit records.

the FROM portion of trigger would be: 
```mysql
    FROM settings s
    WHERE s.key = 'AUDIT_TRIGGER_ENABLE' AND s.value = 'TRUE';
```

chdietz: @malaffoon @agorina let me know what you think about this key and value before I do it, please :)

agorina: Looks good to me.

warehouse/sql/V1_1_0_6__exam_audit.sql
```diff
@@ -0,0 +1,281 @@
...
+-- 2: sample turn on and off
+-- 2: modify audit triggers to respect the on/off switch
+
+USE ${schemaName};
+
+/*
+ Create the setting table that holds setting name and value pairs.
+*/
+CREATE TABLE setting (
+  name VARCHAR(20),
```
chdietz: Should have unique key here I think, will change in next set of audit work.

PR #171, opened 2017-10-16
save work on redshift schema tasks

malaffoon:
I'm just preserving some local changes i made cause i don't plan on working on this for a day or two.

It actually works fine if you set the redshift_* and database_* parameters.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,23 +36,37 @@ def flywayActions = [migrate : org.flywaydb.gradle.task.FlywayMigrateTask,
 
 def reportingSchemas = ["reporting"]
 def warehouseSchemas = ["warehouse"]
-def migrateOlapSchema = ["migrate_olap"]
+def migrateOlapSchemas = ["migrate_olap"]
+def reportingOlapSchemas = ["reporting_olap"]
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -36,23 +36,37 @@ def flywayActions = [migrate : org.flywaydb.gradle.task.FlywayMigrateTask,
...
-                group = "$action$type"
-                url = database_url
-                user = database_user
-                password = database_password
-                driver = 'com.mysql.jdbc.Driver'
-                schemas = ["$schema$type"]
+                def realSchema = schema
                 locations = ['filesystem:' + rootDir.path + "/$schema/sql"]
-                placeholders = ["schemaName": "$schema$type"]
+                if (schema == 'reporting_olap') {
```
malaffoon: The two things this does: 
# Use the configured schema name for redshift
# Don't include redshift tasks in "all" tasks (for now)

wtritch:
I've been working with this during the CI testing/redshift user creation/configuration.  I think it's ready to be merged, unless there are any objections?  FYI @malaffoon @agorina

PR #172, opened 2017-10-16
Add import_id column to root entity tables in reporting_olap schema

wtritch:
This PR is related to https://github.com/SmarterApp/RDW_Ingest/pull/369 which has the associated staging/migration changes

### DIFF
--- a/reporting/sql/V1_1_0_5__rename_import_id.sql
+++ b/reporting/sql/V1_1_0_5__rename_import_id.sql
@@ -1,0 +1,33 @@
+-- Rename import_id columns to update_import_id to represent the warehouse value being migrated
+
+USE ${schemaName};
+
+ALTER TABLE asmt
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE school
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE student
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE student_group
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE exam
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE staging_school
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE staging_student
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE staging_student_group
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE staging_asmt
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE staging_exam
+  CHANGE import_id update_import_id BIGINT NOT NULL;
--- a/reporting_olap/sql/V1_0_0_0__ddl.sql
+++ b/reporting_olap/sql/V1_0_0_0__ddl.sql
@@ -40,7 +40,8 @@ CREATE TABLE staging_asmt (
   school_year smallint NOT NULL,
   name character varying(250) NOT NULL,
   deleted boolean NOT NULL,
-  migrate_id bigint NOT NULL
+  migrate_id bigint NOT NULL,
+  import_id bigint NOT NULL
  );
 
 CREATE TABLE staging_district (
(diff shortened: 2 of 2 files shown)


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -40,7 +40,8 @@ CREATE TABLE staging_asmt (
   school_year smallint NOT NULL,
```
wtritch: The associate V1_0_0_1__dml.sql files starts with `SET SEARCH_PATH to ${schemaName};`  should this file also start by identifying the target schema?

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -54,7 +55,8 @@ CREATE TABLE staging_school (
   district_id int NOT NULL,
   name character varying(100) NOT NULL,
   deleted boolean NOT NULL,
-  migrate_id bigint NOT NULL
+  migrate_id bigint NOT NULL,
+  import_id bigint NOT NULL
```
agorina: I do not think we need it here. Is it there in the `reporting` schema?

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -93,7 +96,8 @@ CREATE TABLE staging_exam (
   scale_score_std_err float,
   performance_level smallint,
   deleted boolean NOT NULL,
-  migrate_id bigint NOT NULL
+  migrate_id bigint NOT NULL,
+  import_id bigint NOT NULL
```
agorina: And I do not think we need it here.

reporting_olap/sql/V1_0_0_0__ddl.sql
malaffoon: I guess it doesn't matter but `district` isn't a master table, do we need/want migrate_id and import_id here? What are the rules for when we want them?

wtritch: I assumed all of that was already decided and just duplicated the existing warehouse/reporting import_id migrations:
school
student
assessment
exam

agorina: I had the same comment, but had to remove it. The way git shows it is confusing, but it is not on the `district`.

malaffoon: Oh. This was my bad reading the deltas ... it looked like the fields (migrate_id and import_id) were in the district table, not the school table.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -150,7 +155,8 @@ CREATE TABLE school (
   id integer encode raw PRIMARY KEY SORTKEY,
   name varchar(100) NOT NULL,
   district_id integer NOT NULL,
-  migrate_id bigint encode delta NOT NULL
+  migrate_id bigint encode delta NOT NULL,
+  import_id bigint encode delta NOT NULL
```
chdietz: Similar to comment above these exam child tables, should they have import_id ?

wtritch: Up to now, we have only recorded import_id on root entities rather than down the entire child table list.  That still seems reasonable to me.  Do you see a need to add it?

wtritch:
Added changes required to modify reporting and reporting_olap migration to record update_import_id rather than import_id

reporting/sql/V1_1_0_5__rename_import_id.sql
```diff
@@ -0,0 +1,33 @@
...
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE staging_student_group
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE staging_asmt
+  CHANGE import_id update_import_id BIGINT NOT NULL;
+
+ALTER TABLE staging_exam
+  CHANGE import_id update_import_id BIGINT NOT NULL;
```
wtritch: I'm just renaming the existing "import_id" column as "update_import_id" which should still be "valid enough" in production since we haven't been processing many, if any, updates. Thoughts? @malaffoon

malaffoon: Seems fine to me.

PR #173, opened 2017-10-17
student_ethnicity audit table and triggers

chdietz:
Edits to V1_1_0_7__student_audit.sql to include the student_ethnicity audit table and triggers.

### DIFF
--- a/warehouse/sql/V1_1_0_7__student_audit.sql
+++ b/warehouse/sql/V1_1_0_7__student_audit.sql
@@ -1,7 +1,7 @@
--- Add
--- 1:
--- todo
--- double check default values on source tables
+-- Add audit tables for student and add a primary key for setting table
+-- 1: Primary key change for setting table
+-- 2: timestamp added to student_ethnicity
+-- 3: student and student_ethnicity audit tables and triggers
 
 USE ${schemaName};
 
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_1_0_7__student_audit.sql
```diff
@@ -11,13 +11,13 @@ USE ${schemaName};
 ALTER TABLE setting ADD PRIMARY KEY (name);
```
chdietz: this table should have primary key so setting can not be duplicated.

warehouse/sql/V1_1_0_7__student_audit.sql
```diff
@@ -11,13 +11,13 @@ USE ${schemaName};
...
-  audited TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) NOT NULL,
-  database_user VARCHAR(255) NOT NULL,
-);
+  Add created timestamp to child tables of parent being audited.
+  Audit records are created for update and delete.
+  Timestamp is required for student updates that include creating a new child record.
+  Child records do not have the import id.
 */
+ALTER TABLE student_ethnicity
+  ADD COLUMN created TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) NOT NULL;
```
chdietz: Same pattern as child tables for exam so we know when the update that is an insert happened.

PR #174, opened 2017-10-17
Add completed_at column to the olap exam fact and staging table.

wtritch:
This PR adds completed_at to the exam staging and fact tables and changes the exam fact table boolean properties to BOOLEAN datatypes.


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -260,6 +262,7 @@ CREATE TABLE fact_student_ica_exam_for_longitudinal (
   claim4_scale_score float encode bytedict,
   claim4_scale_score_std_err float encode bytedict,
   claim4_category smallint encode lzo,
+  completed_at timestamp without time zone encode lzo NOT NULL,
```
wtritch: After running an ANALYZE command on the fact table it came back with an LZO encryption... but we may want to run it again against a larger dataset in the future.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -232,17 +234,17 @@ CREATE TABLE fact_student_ica_exam (
...
-  iep smallint encode lzo NOT NULL,
-  lep smallint encode lzo NOT NULL,
-  section504 smallint encode lzo,
-  economic_disadvantage smallint encode lzo NOT NULL,
-  migrant_status smallint encode lzo,
+  iep boolean encode raw NOT NULL,
+  lep boolean encode raw NOT NULL,
+  section504 boolean encode raw,
+  economic_disadvantage boolean encode raw NOT NULL,
+  migrant_status boolean encode raw,
```
wtritch: On a good note, Redshift performs auto-conversion from smallint to boolean where 0 -> false, !0 -> true so we don't need any special conversion handling.

PR #175, opened 2017-10-18
Feature/add completed at to olap exam schema

wtritch:


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -262,7 +262,7 @@ CREATE TABLE fact_student_ica_exam_for_longitudinal (
   claim4_scale_score float encode bytedict,
   claim4_scale_score_std_err float encode bytedict,
   claim4_category smallint encode lzo,
-  completed_at timestamp without time zone encode lzo NOT NULL,
+  completed_at timestamptz encode lzo NOT NULL,
```
wtritch: Adding timezone information prevents the driver from assuming the DB value is in the local timezone (PDT.)

PR #176, opened 2017-10-18
DWR-1092 Adding more data elements from TRT to warehouse.

agorina:


warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
```diff
@@ -0,0 +1,107 @@
...
+    ADD COLUMN test_window_id varchar(50),
+    ADD COLUMN test_administrator_id varchar(128),
+    ADD COLUMN responsible_organization_name varchar(60),
+    ADD COLUMN test_administrator_name varchar(128),
+    ADD COLUMN session_platform_user_agent varchar(512),
+    ADD COLUMN test_delivery_server varchar(128),
+    ADD COLUMN test_delivery_db varchar(128),
+    ADD COLUMN window_opportunity_count varchar(8),
+    ADD COLUMN theta_score float,
+    ADD COLUMN theta_score_std_err float;
```
agorina: @malaffoon, as we have discussed, I will compile a list of the `varchars` to ask ETS about the size.

warehouse/sql/V1_1_0_9__update_exam_audit_with_more_data.sql
```diff
@@ -0,0 +1,370 @@
+-- modify triggers to add newly added data elements
```
agorina: I broke it into two scripts, just for the readability.

warehouse/sql/V1_1_0_9__update_exam_audit_with_more_data.sql
```diff
@@ -0,0 +1,370 @@
...
+      OLD.response_content_type,
+      OLD.client_id,
+      OLD.page_number,
+      OLD.page_visits,
+      OLD.page_time,
+      OLD.response_type_id
+    FROM setting s
+    WHERE s.name = 'AUDIT_TRIGGER_ENABLE' AND s.value = 'TRUE';
+
+UPDATE setting SET value = 'TRUE' WHERE name = 'AUDIT_TRIGGER_ENABLE';
```
agorina: Since there is currently no ITs to verify that my changes did not break anything I am tempted to remove this like and have the triggers turned off until we have all the tests in place.
@jtreuting , @chdietz , and @malaffoon - please let me know what you think.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
```diff
@@ -0,0 +1,107 @@
...
+  id tinyint NOT NULL PRIMARY KEY,
+  code varchar(10) NOT NULL UNIQUE
+);
+
+INSERT INTO response_type (id, code) VALUES
+  (1, 'value'),
+  (2, 'reference');
+
+ALTER TABLE item
+    ADD COLUMN fiel_test tinyint,
```
malaffoon: `field_test`?

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
```diff
@@ -0,0 +1,107 @@
...
+    ADD COLUMN theta_score_std_err float;
+
+ALTER TABLE exam_item
+    ADD COLUMN administered_at timestamp(0),
+    ADD COLUMN submitted tinyint,
+    ADD COLUMN submited_at timestamp(0),
+    ADD COLUMN number_of_visits smallint,
+    ADD COLUMN response_duration float,
+    ADD COLUMN response_content_type varchar(50),
+    ADD COLUMN client_id varchar(80),
```
malaffoon: `client_id` is by item?  Like it changes in the middle of a test?

warehouse/sql/V1_1_0_9__update_exam_audit_with_more_data.sql
```diff
@@ -0,0 +1,370 @@
+-- modify triggers to add newly added data elements
+USE ${schemaName};
+
+UPDATE setting SET value = 'FALSE' WHERE name = 'AUDIT_TRIGGER_ENABLE';
+
+/*
```
malaffoon: C-style comments in SQL scripts is not consistent

warehouse/sql/V1_1_0_9__update_exam_audit_with_more_data.sql
malaffoon: Believe it or not, i don't have an opinion on this.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
agorina: Thank you, fixing.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
agorina: From TRT spec:"How client identifies the item". Better name?

warehouse/sql/V1_1_0_9__update_exam_audit_with_more_data.sql
agorina: Oh, I copied that script and did not pay attention. Will adjust.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
malaffoon: So a client unique id for every item administered? Wow, okay fine as is.

jtreuting: i know at least for the open source TDS, you can only have a single client defined for the entire instance of TDS.  What do the values look like?  I'm not sure I understand what "How the client identifies the item" means.

warehouse/sql/V1_1_0_9__update_exam_audit_with_more_data.sql
jtreuting: @malaffoon Are you feeling okay?  What is the harm in leaving it?  I'm guessing you are thinking about when/if we push the Update/Delete API changes if schema needs to be pushed as well.  If that is the case then I agree with being cautious.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
```diff
@@ -0,0 +1,107 @@
...
+ALTER TABLE exam
+    ADD COLUMN examinee_id bigint,
+    ADD COLUMN deliver_mode varchar(10),
+    ADD COLUMN hand_score_project int,
+    ADD COLUMN contract varchar(100),
+    ADD COLUMN test_reason varchar(255),
+    ADD COLUMN assessment_admin_started_at date,
+    ADD COLUMN started_at timestamp(0),
+    ADD COLUMN force_submitted_at timestamp(0),
+    MODIFY COLUMN status_date TIMESTAMP(0),
```
jtreuting: super nit: but the all caps TIMESTAMP stood out since it is different from the rest.

warehouse/sql/V1_1_0_9__update_exam_audit_with_more_data.sql
agorina: Here is a potential harm: I have no way to easily test that my changes to the code work. One simple typo in the SQL script and it will break. Since @chdietz is working on setting up the integration tests I feel that it would be better to have this feature off for now.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
agorina: The values look all empty in the samples I have. Maybe we could ask ETS or look up some data in production. 

Just to note that there is also 'clientName' that is part of the opportunity. And it is defined as "Name of the client this test is for (e.g. Oregon).".  The value for this field in the samples I have is "SBAC".

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
agorina: Thank you, fixing.

warehouse/sql/V1_1_0_9__update_exam_audit_with_more_data.sql
chdietz: I don't see the harm in leaving it off, IT's should be done in a day or two.
Then we can turn it on.  If we do leave it on however, existing IT's are exercising triggers because there are IT's for update and they are hitting the triggers and they are running.
The IT's will check that fields align and audit records are inserted  at the correct time.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
chdietz: On this subject are you using Intellij to format SQL.  I have tried and there are some things it does I do not like such as expanding case statements to new lines when I don't want it.  But I think the default behavior is to uppercase types.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
chdietz: Is there a list of elements or attributes we decided not to include for some reason or do we have 'everything'

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
chdietz: The TRDD has this section:
  ```
Appendix B: Omitted Fields: Fields in Test Results Transmission Format (TRT) but not in the data model.
  Item/clientId (optional)
  Item/scoreStatus
  Item/strand
  Item/contentLevel
  Item/pageNumber
  Item/pageVisits
  Item/pageTime
```
So maybe we should not add these?

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
agorina: We have not included everything, but I did not compile a list of what was excluded. I did review it with @malaffoon and @jtreuting, but if you want an overview, please let me know.

agorina:
I think we should include those data elements, but open to other suggestions. @jtreuting  and @malaffoon, please share your thoughts on this subject.

warehouse/sql/V1_1_0_8__add_more_data_from_trt.sql
malaffoon: What i was thinking when Alla and i discussed these fields: we don't know how this stuff will be used in the future. We've been (slightly) burned by not capturing fields that we were "sure would never be needed in reporting." And disk space is cheap. And we have the separation between the warehouse and the data mart. So, why not capture the data, then use what we need now, keeping the other data around for future work/features or complex analytics. 

Even after these changes there will be some data we don't capture, why are we treating those differently? Well, most of those require a non-trivial data model change which would have too great an impact at this point in the project. So there is a line.

chdietz: ok, just wanted to see if that fact these are mentioned the way they are in the TRDD was discussed.  It caught my attention when reading about `client_id`.
If it turns out we do not need them we can always remove later.

jtreuting:
@agorina and @chdietz, reading the comment thread I'm having a hard time following which fields you all are discussing adding that aren't there.  Maybe someone could list those explicitly as there are some shown above but then followed by "so maybe we shouldn't add these."  So I am a bit confused :) .

jtreuting:
@agorina clarified the fields being discussed.  yeah i think they should be included as they can provide value and I can come up with a hypothetical as to why having that data would be valuable and could come up for data analysis later on

PR #178, opened 2017-10-20
DWR-1092 Remove exam item client id.

agorina:
ETS confirmed that always receive blank data from AIR in this field and it is not in the data model.


chdietz:
@jtreuting Yes. Of course in IDE you can pull up the ddl of the current trigger or look for the last flyway where that trigger was re-created as a start point.  Also the way I am writing the IT's if a column is left off the audit table or a column is left of the trigger sql they will catch it.

PR #179, opened 2017-10-22
DWR-1104 Create a view that represents a student current school.

agorina:
This is just to test the idea. 
Please let me know what you think about this approach for determining a "current institution" for a student record. 

The idea is that this view will be used by migrate(s). Both migrates could use it if needed. 
If at some point SBAC decides to use a true roster, this view could be replaced with a table.

Of course performance is a concern. From my preliminary reading it looks like views in MySQL should perform no worse than just a query. And if we run into issues, we can then decide to create a permanent table with the same data elements.


agorina:
This query was recommended by Google as the common solution to the given problem. I take no credit or responsibility for it :)

malaffoon:
Seems good.

PR #180, opened 2017-10-24
DWR-1099 Update is called on warehouse.student when not actually updated

agorina:
When student is `upserted` the ethnicity may change too.
Ethnicity is handled in java outside of the stored procedure. 
This change returns a flag indicating if the import id has been changed so that when we modify ethnicity in java we know to update or not import id.

### DIFF
--- a/warehouse/sql/V1_1_0_12__modify_student_upsert_to_return_action.sql
+++ b/warehouse/sql/V1_1_0_12__modify_student_upsert_to_return_action.sql
@@ -1,0 +1,66 @@
+-- modify student upsert to return an action
+
+USE ${schemaName};
+
+/************************************* Stored procedures ***************************************/
+
+DROP PROCEDURE IF EXISTS student_upsert;
+DELIMITER //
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_1_0_12__modify_student_upsert_to_return_action.sql
```diff
@@ -0,0 +1,66 @@
...
+                                IN  p_first_name                    VARCHAR(60),
+                                IN  p_middle_name                   VARCHAR(60),
+                                IN  p_gender_id                     TINYINT,
+                                IN  p_first_entry_into_us_school_at DATE,
+                                IN  p_lep_entry_at                  DATE,
+                                IN  p_lep_exit_at                   DATE,
+                                IN  p_birthday                      DATE,
+                                IN  p_import_id                     BIGINT,
+                                OUT p_id                            INT,
+                                OUT p_updated                       TINYINT)
```
chdietz: You added an indication if there was an update, correct?

agorina: Yes, that is correct.

PR #182, opened 2017-10-25
DWR-1104 Store "current institution" for a student record

agorina:
This includes all the changes.


warehouse/sql/V1_1_0_13__student_current_school.sql
```diff
@@ -0,0 +1,87 @@
...
+                                IN  p_last_or_surname               VARCHAR(60),
+                                IN  p_first_name                    VARCHAR(60),
+                                IN  p_middle_name                   VARCHAR(60),
+                                IN  p_gender_id                     TINYINT,
+                                IN  p_first_entry_into_us_school_at DATE,
+                                IN  p_lep_entry_at                  DATE,
+                                IN  p_lep_exit_at                   DATE,
+                                IN  p_birthday                      DATE,
+                                IN  p_inferred_school_id            INT,
+                                IN  p_inferred_effective_date       TIMESTAMP(6),
```
agorina: Added the above two parameters.

warehouse/sql/V1_1_0_13__student_current_school.sql
```diff
@@ -0,0 +1,87 @@
...
+    THEN
+       IF(p_inferred_school_id IS NULL)
+         THEN
+            SELECT e1.school_id, e1.completed_at INTO p_inferred_school_id, p_inferred_effective_date
+                FROM (SELECT * FROM exam WHERE deleted = 0 and student_id = p_id) AS e1
+                    LEFT OUTER JOIN (SELECT * FROM exam WHERE deleted = 0 and student_id = p_id) AS e2
+                    ON e1.student_id = e2.student_id AND (e1.completed_at < e2.completed_at
+                    OR (e1.completed_at = e2.completed_at AND e1.Id < e2.Id))
+                WHERE e2.student_id IS NULL;
+       END IF;
```
agorina: Added the above IF block.

warehouse/sql/V1_1_0_13__student_current_school.sql
```diff
@@ -0,0 +1,87 @@
...
+            AND first_name <=> p_first_name
+            AND middle_name <=> p_middle_name
+            AND gender_id <=> p_gender_id
+            AND first_entry_into_us_school_at <=> p_first_entry_into_us_school_at
+            AND lep_entry_at <=> p_lep_entry_at
+            AND lep_exit_at <=> p_lep_exit_at
+            AND birthday <=>  p_birthday
+            AND (inferred_effective_date <=> p_inferred_effective_date OR (inferred_effective_date IS NOT NULL AND inferred_effective_date > COALESCE(p_inferred_effective_date,inferred_effective_date)))
+            AND (inferred_school_id <=> p_inferred_school_id OR (inferred_effective_date IS NOT NULL AND inferred_effective_date > COALESCE(p_inferred_effective_date,inferred_effective_date)));
+
```
agorina: Added the above two ANDs.

warehouse/sql/V1_1_0_13__student_current_school.sql
```diff
@@ -0,0 +1,87 @@
...
+          last_or_surname               = p_last_or_surname,
+          first_name                    = p_first_name,
+          middle_name                   = p_middle_name,
+          gender_id                     = p_gender_id,
+          first_entry_into_us_school_at = p_first_entry_into_us_school_at,
+          lep_entry_at                  = p_lep_entry_at,
+          lep_exit_at                   = p_lep_exit_at,
+          birthday                      = p_birthday,
+          inferred_effective_date       = p_inferred_effective_date,
+          inferred_school_id            = p_inferred_school_id,
```
agorina: Added the two lines above.

warehouse/sql/V1_1_0_13__student_current_school.sql
```diff
@@ -0,0 +1,87 @@
...
+          update_import_id              = p_import_id
+        WHERE id = p_id;
+      END IF;
+    ELSE
+      INSERT INTO student (ssid, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at, lep_entry_at, lep_exit_at, birthday, inferred_school_id, inferred_effective_date, import_id, update_import_id)
+      VALUES (p_ssid, p_last_or_surname, p_first_name, p_middle_name, p_gender_id, p_first_entry_into_us_school_at, p_lep_entry_at, p_lep_exit_at, p_birthday, p_inferred_school_id, p_inferred_effective_date, p_import_id, p_import_id);
+
```
agorina: Added two new fields for insert.

warehouse/sql/V1_1_0_13__student_current_school.sql
```diff
@@ -0,0 +1,87 @@
...
+  ADD COLUMN inferred_school_id int,
+  ADD INDEX idx__student__inferred_school (inferred_school_id),
+  ADD CONSTRAINT fk__student__inferred_school FOREIGN KEY (inferred_school_id) REFERENCES school(id);
+
+
+/************************************* Stored procedures ***************************************/
+
+DROP PROCEDURE IF EXISTS student_upsert;
+DELIMITER //
+CREATE PROCEDURE student_upsert(IN  p_ssid                          VARCHAR(65),
```
agorina: This is one ugly stored procedure...but...

reporting/sql/V1_1_0_6__student_current_school.sql
```diff
@@ -0,0 +1,9 @@
+-- add student current school/district and exam ids from which it was derived
+USE ${schemaName};
+
+ALTER TABLE student
+    ADD COLUMN inferred_school_id int,
+    ADD CONSTRAINT fk__student__inferred_school_id FOREIGN KEY (inferred_school_id) REFERENCES school(id);
+
+ALTER TABLE staging_student
+    ADD COLUMN inferred_school_id int;
```
wtritch: FYI, I just stole V1_1_0_6 =(

warehouse/sql/V1_1_0_13__student_current_school.sql
```diff
@@ -0,0 +1,87 @@
+USE ${schemaName};
+
+ALTER TABLE student
+  ADD COLUMN inferred_effective_date TIMESTAMP(6) DEFAULT NULL,
+  ADD COLUMN inferred_school_id int,
+  ADD INDEX idx__student__inferred_school (inferred_school_id),
+  ADD CONSTRAINT fk__student__inferred_school FOREIGN KEY (inferred_school_id) REFERENCES school(id);
```
wtritch: A foreign key still requires a separate index?

reporting/sql/V1_1_0_6__student_current_school.sql
agorina: Oh, thank you - will change.

warehouse/sql/V1_1_0_13__student_current_school.sql
agorina: Yeah, we have learned that it is easier to manage indexes this way with MySQL.

warehouse/sql/V1_1_0_13__student_current_school.sql
wtritch: What if it's an `update` request and the completed_at date has changed?

warehouse/sql/V1_1_0_13__student_current_school.sql
wtritch: maybe I'm missing something, but should those be LESS-THANs rather than GREATER-THANs?

warehouse/sql/V1_1_0_13__student_current_school.sql
wtritch: The various SQL flavors give me a headache...

warehouse/sql/V1_1_0_13__student_current_school.sql
agorina: This is what I am trying to do (still adding tests to prove that it works, so not sure if this code does it ) : 
If TRT is 'insert` or `update`, then we take the data from there and pass in. 
If the TRT school is different from the current inferred school and the TRT date is after the 'inferred school date`, then it gets updated.

warehouse/sql/V1_1_0_13__student_current_school.sql
agorina: This trying to decide if to update - so the logic is opposite.

wtritch: ah, got it.  True == skip

warehouse/sql/V1_1_0_13__student_current_school.sql
chdietz: @wtritch try thinking in sets ;)

warehouse/sql/V1_1_0_13__student_current_school.sql
chdietz: The logic looks ok but this SP is getting complex.  We have an exam inferring a student upsert and in that logic to infer the current registration.

warehouse/sql/V1_1_0_13__student_current_school.sql
agorina: Are you mocking me? :)

warehouse/sql/V1_1_0_13__student_current_school.sql
```diff
@@ -0,0 +1,138 @@
...
+      END IF;
+      LEAVE partition_loop;
+    END LOOP partition_loop;
+
+  END;
+//
+DELIMITER ;
+
+-- NOTE that this does not account for DELETED exams assuming that we do not have a case when deleted exam has the most recent completed_at
+CALL loop_by_partition(
```
agorina: This is pretty slow, that is why I am not including deleted. 
Any concerns?

jtreuting: This is for the first time migration where we need to populate current school for all students right

agorina: Yes, this is correct.

agorina: Oh, I just realized that this need to be changed. I need to add an import record and make sure students  get updated so that they are re-migrated. Will do the change.

warehouse/sql/V1_1_0_14__initial_load_of_current_school.sql
```diff
@@ -0,0 +1,83 @@
+USE ${schemaName};
+
+ALTER TABLE student
+  MODIFY COLUMN updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
```
agorina: The `updated` will be reset manually later in the scripts. 
This is done to spread the timestamps to contain the size of the chunks during migrate. It uses the same approach as the historical data load. Otherwise we end up with too many records and reporting migrate runs out of memory.

warehouse/sql/V1_1_0_14__initial_load_of_current_school.sql
```diff
@@ -0,0 +1,83 @@
...
+        ITERATE partition_loop;
+      END IF;
+      LEAVE partition_loop;
+    END LOOP partition_loop;
+
+  END;
+//
+DELIMITER ;
+
+-- NOTE that this does not account for DELETED exams assuming that we do not have a case when deleted exam has the most recent completed_at
```
agorina: Wanted to point this out.

malaffoon: Is it a lot slower when you account for deleted exams?

agorina: It is significantly slower when I account for deletes.
If you think we should consider them, I can play with it more. I thought that for what we have now it is not important.

And I forgot to mention that now it takes about 10 seconds per each partition on aws dev. The whole process is about 30+ min.

jtreuting: For the assumption about the deletes I think it depends on if we can push this change to set the current school for students before we push the feature that relies on it, namely the student transfer access.  Otherwise we will be pushing the DELETE API before we run this and I'm not sure we can make this assumption safely.

agorina: Discussed offline, but adding a note here for people who may follow this review: for now we assume that this will be released together with the DELETE API. If this does not work out this way, then we will need to change the approach here.

PR #183, opened 2017-10-27
DWR-1067 Store additional item data

agorina:


reporting/sql/V1_1_0_8__additional_item_info.sql
```diff
@@ -0,0 +1,15 @@
+USE ${schemaName};
+
+ALTER TABLE item
+    ADD COLUMN field_test tinyint,
+    ADD COLUMN active tinyint,
+    ADD COLUMN type varchar(40),
+    ADD COLUMN answer_options tinyint,
+    ADD COLUMN answer text;
```
agorina: Is this the right data type? Can we make is smaller?

wtritch: I think text is right... some of the answers are biggish xml blobs... =/

agorina: Per @jtreuting it is not the full answer, it is just the key. So changed the type and name.

reporting/sql/V1_1_0_8__additional_item_info.sql
```diff
@@ -0,0 +1,15 @@
+USE ${schemaName};
+
+ALTER TABLE item
+    ADD COLUMN field_test tinyint,
+    ADD COLUMN active tinyint,
+    ADD COLUMN type varchar(40),
+    ADD COLUMN answer_options tinyint,
```
jtreuting: is it clear that this is the number of options and not the actual options?  i know it is when looking at the datatype since it's tinyint and not varchar, but at a glance maybe it isn't.  `number_of_answer_options`?? (maybe too verbose?)

agorina: Will rename to number_of_answer_options.

reporting/sql/V1_1_0_8__additional_item_info.sql
```diff
@@ -0,0 +1,15 @@
+USE ${schemaName};
+
+ALTER TABLE item
+    ADD COLUMN field_test tinyint,
+    ADD COLUMN active tinyint,
+    ADD COLUMN type varchar(40),
+    ADD COLUMN number_of_answer_options tinyint,
```
malaffoon: What is this? Is this a count of how many answer options there are? So it would be 6 for A-F?

jtreuting: correct.  we don't care about the actual text of the answer options, and they don't store A, B, C with the actual option text.  every system uses the convention of starting with A and going thru for each option.  That is what TDS does for display as well as IRIS and how IAT is handling it moving forward.

agorina: I am not sure I know the answer, but looking at the sample file I have values 0 through 6. @jtreuting, please correct me if needed.

reporting/sql/V1_1_0_8__additional_item_info.sql
```diff
@@ -0,0 +1,15 @@
...
+    ADD COLUMN answer_key varchar(50);
```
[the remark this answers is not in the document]
agorina: I do not think we have answer_key in the logical data model, do we?
I got this size from @jtreuting, so maybe he could help me here (do not mean to finger-point...)

reporting/sql/V1_1_0_8__additional_item_info.sql
malaffoon: I think that column name is going to be annoying over time. `options_count`, `num_options`?

reporting/sql/V1_1_0_8__additional_item_info.sql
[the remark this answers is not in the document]
jtreuting: i have only seen the comma delimited and text that says something like "Hand" (for hand scoring) or "qrx" (for machine scored logic in separate file).  So that 50 came from me.  Maybe the IRP that I ran is not representative?  The long text is from the rubric and exemplar that is displayed in the UI and comes from a different field.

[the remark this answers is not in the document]
jtreuting: thanks @agorina "hey jeff ... here is a bus" :)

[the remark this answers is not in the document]
malaffoon: I'm probably confusing answer_key with item-response data.

reporting/sql/V1_1_0_8__additional_item_info.sql
agorina: Ok, renaming (once again) to `options_count`.

reporting/sql/V1_1_0_8__additional_item_info.sql
jtreuting: Students give a full response but that is the student response and not the answer key for an item.  The answer key seems to be "A,C" for multiple choice or multiple select, it says "Hand" for hand scoring and "qrx" for machine scored items that have the rules in a QRX file that is separate.

reporting/sql/V1_1_0_8__additional_item_info.sql
jtreuting: they said that in practice they have items with up to 7 options.  Our code can handle A-Z and that is the same approach IAT takes.

PR #184, opened 2017-11-03
Add missing created columns to child audit tables

chdietz:
created column found missing in audit exam child tables.  There is an update to IT's after this PR is merged.

### DIFF
--- a/warehouse/sql/V1_1_0_16__add_audit_columns.sql
+++ b/warehouse/sql/V1_1_0_16__add_audit_columns.sql
@@ -1,0 +1,219 @@
+-- Add missing columns to audit exam child tables
+-- 1: add the created column that comes from the created timestamp on the exam_ child record
+-- 2: re create the triggers
+-- 3: this will coincide with integration tests updates on these tables that ensure all columns are audited.
+
+USE ${schemaName};
+
+-- Add created columns to audit tables.
+-- Since auditing is not yet released these tables should have rows deleted.
+-- This truncate should be removed when production release combined flyway is created.
+TRUNCATE TABLE audit_exam_claim_score;
+TRUNCATE TABLE audit_exam_available_accommodation;
+TRUNCATE TABLE audit_exam_item;
+
+ALTER TABLE audit_exam_claim_score
+  ADD COLUMN created TIMESTAMP(6) NOT NULL;
+
+ALTER TABLE audit_exam_available_accommodation
+  ADD COLUMN created TIMESTAMP(6) NOT NULL;
+
+ALTER TABLE audit_exam_item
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_1_0_16__add_audit_columns.sql
```diff
@@ -0,0 +1,219 @@
...
+
+USE ${schemaName};
+
+-- Add created columns to audit tables.
+-- Since auditing is not yet released these tables should have rows deleted.
+-- This truncate should be removed when production release combined flyway is created.
+TRUNCATE TABLE audit_exam_claim_score;
+TRUNCATE TABLE audit_exam_available_accommodation;
+TRUNCATE TABLE audit_exam_item;
+
```
chdietz: I had conversations about this will multiple people and arrived at this compromise.

warehouse/sql/V1_1_0_16__add_audit_columns.sql
```diff
@@ -0,0 +1,219 @@
...
+      OLD.exam_id,
+      OLD.subject_claim_score_id,
+      OLD.scale_score,
+      OLD.scale_score_std_err,
+      OLD.category,
+      OLD.theta_score,
+      OLD.theta_score_std_err,
+      OLD.created
+    FROM setting s
+    WHERE s.name = 'AUDIT_TRIGGER_ENABLE' AND s.value = 'TRUE';
```
malaffoon: I didn't comment on this in the original commits but ...

For the setting, would it be more readable to use guard-style coding in this trigger? You could have a helper function that returns whether to audit or not and then exit the trigger if that returns false. Then the insert is a simple VALUES instead of a SELECT. Just musing ...

chdietz: Interesting idea because with this style I have now the insert always runs, it's just that there is nothing to insert if the setting is not 'TRUE'.

I merged this now because I have integration test changes that address todo's in ExamRepositoryIT.

I noticed that with the style I have now if there is a problem with the insert in the trigger you see the error even if the trigger is setting to not TRUE.

Another aspect is with your suggestion the helper function would be an SP correct? so the 
`FROM setting s WHERE s.name = 'AUDIT_TRIGGER_ENABLE' AND s.value = 'TRUE'`
would only have to defined once. I think I like this aspect.

If we want to do this I would do all triggers in one flyway again and then the additional for groups tables would follow the same pattern.

agorina: The helper could be a VIEW.
I think it may be simpler this way, but I am sure there is more than one option.

PR #185, opened 2017-11-06
modify completed_at to retain nanosecond precision

malaffoon:
@agorina is there anything to do for redshift on this? It didn't look like it to me.


chdietz:
Does this need to be changed anywhere in olap?

PR #187, opened 2017-11-08
DWR-1104  Fix a bug with the inferred school

agorina:


warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,81 @@
...
+          middle_name                   = p_middle_name,
+          gender_id                     = p_gender_id,
+          first_entry_into_us_school_at = p_first_entry_into_us_school_at,
+          lep_entry_at                  = p_lep_entry_at,
+          lep_exit_at                   = p_lep_exit_at,
+          birthday                      = p_birthday,
+          inferred_effective_date       = CASE WHEN (p_inferred_effective_date IS NULL OR p_inferred_effective_date >= COALESCE(inferred_effective_date, p_inferred_effective_date)) THEN p_inferred_effective_date ELSE inferred_effective_date END,
+          inferred_school_id            = CASE WHEN (p_inferred_effective_date IS NULL OR p_inferred_effective_date >= COALESCE(inferred_effective_date, p_inferred_effective_date)) THEN p_inferred_school_id ELSE inferred_school_id END,
+          update_import_id              = p_import_id
+        WHERE id = p_id;
```
agorina: Lines changed are 69-71, I added a CASE statement to not update the inferred school/date IF the given new inferred date is before the existing one.

Here is a use case:
A user has an inferred school/date, an exam with the older completed date AND changes to the student core attributes is submitted. This was falling through as an update and changed the school/date to the old one.
In this case we should not update any of the student's attributes, but we have discussed that due to the fact that this is very unlikely to happen and if it does the next exam will correct the data we decided to keep it this way.

chdietz: Does this address the issue new exam, new completedDate and nothing else changed causing a migrate and audit?

agorina: No, it does not address that issue. I am not sure if we should group the solution into one though.

chdietz: In any case the new logic looks good

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,81 @@
...
+        SET
+          last_or_surname               = p_last_or_surname,
+          first_name                    = p_first_name,
+          middle_name                   = p_middle_name,
+          gender_id                     = p_gender_id,
+          first_entry_into_us_school_at = p_first_entry_into_us_school_at,
+          lep_entry_at                  = p_lep_entry_at,
+          lep_exit_at                   = p_lep_exit_at,
+          birthday                      = p_birthday,
+          inferred_effective_date       = CASE WHEN (p_inferred_effective_date IS NULL OR p_inferred_effective_date >= COALESCE(inferred_effective_date, p_inferred_effective_date)) THEN p_inferred_effective_date ELSE inferred_effective_date END,
```
chdietz: not sure I understand this piece:
`p_inferred_effective_date >= COALESCE(inferred_effective_date, p_inferred_effective_date))`

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
agorina: After chatting more with @malaffoon  we have agreed to change the logic and repurpose  `inferred_effective_date` as `student_data_date`. This means that none of the student data will be updated. I am working on the change, please let me know if you have any objections.
@jtreuting , @chdietz .

For now, I am not going to rename inferred_effective_date, but when it is moved into a diff table I will rename it as well.

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
agorina: this will go away  :)

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,92 @@
...
+      SELECT id, 0 INTO p_id, p_updated FROM student WHERE ssid = p_ssid;
+    END;
+
+    -- see if it is an existing student and if the data is fresher than the one stored
+    SELECT id,
+            CASE WHEN (p_inferred_effective_date IS NOT NULL AND inferred_effective_date IS NOT NULL AND p_inferred_effective_date < inferred_effective_date)
+                THEN 0
+            ELSE 1 END
+            INTO p_id, p_updated
+        FROM student WHERE ssid = p_ssid;
```
agorina: This main change is here.

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,92 @@
...
+
+    IF (p_id IS NOT NULL AND p_updated = 1)
```
agorina: And here.

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,92 @@
...
+          first_entry_into_us_school_at = p_first_entry_into_us_school_at,
+          lep_entry_at                  = p_lep_entry_at,
+          lep_exit_at                   = p_lep_exit_at,
+          birthday                      = p_birthday,
+          inferred_effective_date       = p_inferred_effective_date,
+          inferred_school_id            = p_inferred_school_id,
+          update_import_id              = p_import_id
+        WHERE id = p_id;
+      END IF;
+    ELSEIF (p_id IS NULL)
```
agorina: And this else was adjusted.

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
chdietz: so:
1. you will rename the column?
2. maybe get rid of `inferred_school_id` if the other option works out?

Process order question:
If `student_data_date` is checked to see if updatable fields should be updated then this
is this a similar concern to inferred date, meaning we have to update this date even
if there are no changes forcing a migrate/audit because of this use case?

Option: Maybe we say we process student changes in the order we receive them. And a more robust solution is integrate an SIS feed?  If we did not implement student transfer would we even be considering `student_data_date`?

```text
Assume LEPStatus represents all updatable fields and same student
               
Exam LEPStatus completedDate 
---- --------- -------------
  1    Y       1                Processed      
  2    N       2                Not submit yet
  3    Y       3                Processed      
  3    .       .                Processed     
```
Reset Question:
If an exam is reset do we want to update to the most recent non reset exam?  Because a reset could be wrong SSID or something that would make the student data that is updatable completely wrong.

agorina: Let me see if I can try to answer all of your questions :)
1. "you will rename the column?" - I am planning on moving `inferred` data into a separate table and yes, I will rename the column there.
2. "maybe get rid of inferred_school_id if the other option works out?" - not sure what you mean under other option. Are you referring to CALPAD data? Is yes, then we would have a real school id instead. But we will deal with it when it comes.
3. Yes, the issue of updating the date and causing extra auditing is still here. I will work on it later.
4. Reset should be handled now. Reset is a soft-delete and in this case a NULL value is passed in for the `inferred` data and this SP derives them based on the available exams. I have not changed that.

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,92 @@
...
+    SELECT id,
+            CASE WHEN (p_inferred_effective_date IS NOT NULL AND inferred_effective_date IS NOT NULL AND p_inferred_effective_date < inferred_effective_date)
+                THEN 0
+            ELSE 1 END
+            INTO p_id, p_updated
+        FROM student WHERE ssid = p_ssid;
+
+    IF (p_id IS NOT NULL AND p_updated = 1)
+    THEN
+       IF(p_inferred_school_id IS NULL)
```
malaffoon: minor nit: i think it is easier to read with the whitespace and `THEN` on the same line, so
```
    IF (p_inferred_school_id is NULL) THEN
```
(assuming that is legal syntax)

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,92 @@
...
+         THEN
+            SELECT e1.school_id, e1.completed_at INTO p_inferred_school_id, p_inferred_effective_date
```
malaffoon: i know it's fresh in our minds but perhaps add a comment, something like:
```
    -- caller doesn't have school data so infer school from any existing exams
```

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,92 @@
...
+            AND middle_name <=> p_middle_name
+            AND gender_id <=> p_gender_id
+            AND first_entry_into_us_school_at <=> p_first_entry_into_us_school_at
+            AND lep_entry_at <=> p_lep_entry_at
+            AND lep_exit_at <=> p_lep_exit_at
+            AND birthday <=>  p_birthday
+            AND (inferred_effective_date <=> p_inferred_effective_date OR (inferred_effective_date IS NOT NULL AND inferred_effective_date > COALESCE(p_inferred_effective_date,inferred_effective_date)))
+            AND (inferred_school_id <=> p_inferred_school_id OR (inferred_effective_date IS NOT NULL AND inferred_effective_date > COALESCE(p_inferred_effective_date,inferred_effective_date)));
+
+      IF (p_updated = 1)
```
malaffoon: ditto

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,92 @@
...
+          inferred_school_id            = p_inferred_school_id,
+          update_import_id              = p_import_id
+        WHERE id = p_id;
+      END IF;
+    ELSEIF (p_id IS NULL)
+    THEN
+      INSERT INTO student (ssid, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at, lep_entry_at, lep_exit_at, birthday, inferred_school_id, inferred_effective_date, import_id, update_import_id)
+      VALUES (p_ssid, p_last_or_surname, p_first_name, p_middle_name, p_gender_id, p_first_entry_into_us_school_at, p_lep_entry_at, p_lep_exit_at, p_birthday, p_inferred_school_id, p_inferred_effective_date, p_import_id, p_import_id);
+
+      SELECT id, 2 INTO p_id, p_updated FROM student WHERE ssid = p_ssid;
```
malaffoon: perhaps *slightly* more performant to just do:
```
    SELECT LAST_INSERT_ID(), 2 INTO p_id, p_updated;
```

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,92 @@
...
+                                OUT p_id                            INT,
+                                OUT p_updated                       TINYINT)
+  BEGIN
+    --  handle duplicate entry: if there are two competing inserts, one will end up here
+    DECLARE CONTINUE HANDLER FOR 1062
+    BEGIN
+      SELECT id, 0 INTO p_id, p_updated FROM student WHERE ssid = p_ssid;
+    END;
+
+    -- see if it is an existing student and if the data is fresher than the one stored
```
malaffoon: nit: perhaps add to this comment something like:
```
    -- (note that setting p_updated=1 may indicate create or update at this point)
```

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,73 @@
...
+    SELECT id, 0 INTO p_id, p_updated FROM student WHERE ssid = p_ssid;
+
+    IF (p_id IS NOT NULL) THEN
+
+      SELECT CASE WHEN p_inferred_effective_date IS NULL OR completed_at > p_inferred_effective_date THEN school_id ELSE p_inferred_school_id END INTO p_inferred_school_id
+        FROM exam WHERE student_id = p_id AND deleted = 0
+        ORDER BY completed_at DESC LIMIT 1;
```
agorina: @chdietz , @malaffoon and @wtritch :
This is a combination of what @wtritch was suggesting with some additional flavor.
We pass in the date and school from a TRT and compare it with the latest exam. If the given value is newer, it is saved (even though the exam is not stored yet). Otherwise the latest exam's school is taken.

I am going to close this review and create a new one since I have to remove 'inferred_effective_date' from everywhere.  Just wanted to share the solution.

Thank you all for the help!

PR #188, opened 2017-11-10
DWR-1104  Another attempt to fix 'inferred headache'

agorina:


warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,143 @@
+USE ${schemaName};
+
+ALTER TABLE student
+  DROP COLUMN inferred_effective_date;
```
agorina: We do not need to store it anymore.

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,143 @@
...
+      SELECT id, 0 INTO p_id, p_updated FROM student  WHERE ssid = p_ssid;
+    END;
+
+    SELECT id, 0 INTO p_id, p_updated FROM student WHERE ssid = p_ssid;
+
+    IF (p_id IS NOT NULL) THEN
+      -- infer a school based on the given effective date and existing exams
+      SELECT CASE WHEN p_inferred_effective_date IS NULL OR completed_at > p_inferred_effective_date THEN school_id ELSE p_inferred_school_id END INTO p_inferred_school_id
+        FROM exam WHERE student_id = p_id AND deleted = 0
+      ORDER BY completed_at DESC LIMIT 1;
```
agorina: I played with adding an index on `completed_at` (on AWS dev) and it seemed to ignore it and use the primary key. Will need to test it more as part of the performance testing JIRA.

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
```diff
@@ -0,0 +1,143 @@
...
+                                IN  p_last_or_surname               VARCHAR(60),
+                                IN  p_first_name                    VARCHAR(60),
+                                IN  p_middle_name                   VARCHAR(60),
+                                IN  p_gender_id                     TINYINT,
+                                IN  p_first_entry_into_us_school_at DATE,
+                                IN  p_lep_entry_at                  DATE,
+                                IN  p_lep_exit_at                   DATE,
+                                IN  p_birthday                      DATE,
+                                IN  p_inferred_school_id            INT,
+                                IN  p_inferred_effective_date       TIMESTAMP(6),
```
jtreuting: i think i'm missing one piece of the equation.  If we are no longer storing the inferred_effective_date, what is passed in here?

agorina: The passed in values are from the new TRT. This lets us compare the new TRT data with the latest exam data to choose the winner based on the dates.

Note that 'RESETS' are handled differently. With the RESET TRT, we first soft-delete the exam and then pass in NULL values in here.

jtreuting: so is it more accurate to say that this is the completed_at from the TRT.  so should it be named `p_completed_at`?

malaffoon: Jeff's suggestion is good, i think both of these parameters should be renamed as if they are just from the exam. Either just `p_school_id` and `p_completed_at` or maybe, to be a bit more clear `p_exam_school_id` and `p_exam_completed_at` (or maybe `p_new_school_id`, `p_new_completed_at`).

jtreuting: i like the `p_exam_` prefix i think

agorina: Will rename to `p_exam_`. Thank you.

warehouse/sql/V1_1_0_19__student_upsert_fix_inferred_school.sql
chdietz: p_exam_... is easier to read, thanks

PR #189, opened 2017-11-14
Increase the size of the t3_program_type column to account for the lo…

wtritch:
…ngest valid value: ShelteredEnglishInstruction

### DIFF
--- a/reporting/sql/V1_1_0_10__increase_t3_program_type_size.sql
+++ b/reporting/sql/V1_1_0_10__increase_t3_program_type_size.sql
@@ -1,0 +1,9 @@
+-- Increase the size of the t3_program_type column to account for all valid values
+
+USE ${schemaName};
+
+ALTER TABLE exam
+  MODIFY COLUMN t3_program_type varchar(30);
+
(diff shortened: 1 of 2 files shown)


reporting/sql/V1_1_0_10__increase_t3_program_type_size.sql
```diff
@@ -0,0 +1,9 @@
+-- Increase the size of the t3_program_type column to account for all valid values
+
+USE ${schemaName};
+
+ALTER TABLE exam
+  MODIFY COLUMN t3_program_type varchar(30);
+
+ALTER TABLE staging_exam
+  MODIFY COLUMN t3_program_type varchar(30);
```
wtritch: Longest valid value has a length of 27, so I bumped max length to 30.

agorina:
I am assuming that there will be a corresponding Ingest change too (we are validating the length during ingest).

wtritch:
@agorina thank you for catching that!  I almost missed it.  Here's the PR: https://github.com/SmarterApp/RDW_Ingest/pull/410

PR #190, opened 2017-11-18
Feature/validation scripts

esotrope:
First pass at a migration validation script.


validation/validate-reporting-iab.sql
```diff
@@ -0,0 +1,162 @@
...
+    count(*),
+    code
+  FROM exam e
+    JOIN exam_available_accommodation ea ON e.id = ea.exam_id
+    JOIN accommodation a ON a.id = ea.accommodation_id
+  WHERE e.type_id = 2
+  GROUP BY ea.accommodation_id
+  ORDER BY count(*);
+
+# -- Student
```
esotrope: disabled queries outside of iab/ica context for now.

validation/validate-reporting-ica.sql
```diff
@@ -0,0 +1,131 @@
...
+    'asmt year',
+    'admin condition',
+    'complete';
+INSERT INTO ica_validation (testNum, result1, result2, result3, result4, result5)
+  SELECT
+    (SELECT max(testNum)
+     FROM ica_validation),
+    count(*),
+    a.id,
+    a.school_year,
```
esotrope: changed this from `e.school_year` in each script. I think the existing post-validation script incorrectly had it as `e.school_year` correct me if I'm wrong though.

validation/validate-migration.sh
```diff
@@ -0,0 +1,144 @@
...
+script_dir=.
+out_dir=`mktemp -d`
+warehouse_mysql_conf=`mktemp`
+reporting_mysql_conf=`mktemp`
+
+warehouse_ica=${out_dir}/warehouse_ica.csv
+warehouse_iab=${out_dir}/warehouse_iab.csv
+reporting_ica=${out_dir}/reporting_ica.csv
+reporting_iab=${out_dir}/reporting_iab.csv
+reporting_olap_ica=${out_dir}/reporting_olap_ica.csv
```
esotrope: You'll notice i split the script up by asmt type so i could compare apples to apples with reporting & reporting olap output

validation/README.md
```diff
@@ -0,0 +1,81 @@
...
+```bash
+./validate-migration.sh [environment_properties_file]
+```
+#### Output
+
+##### Success
+```
+completed in 00:00:08
+
+/var/folders/h6/gw7zs2cd0cl0v5pd42gtf4740000gp/T/tmp.EdjR0AEZ
```
esotrope: i guess i can change this back to a `results-${timestamp}` formatted dir and just use the gitignore to prevent checkin. thoughts?

validation/validate-migration.sh
```diff
@@ -0,0 +1,144 @@
...
+tree_options="--noreport"
+
+report_query="select testNum, result1, result2, result3, result4, result5 FROM ica_validation ORDER BY testNum, id;"
+
+# set up
+
+mkdir -p ${out_dir}
+
+echo -e "[client]\npassword=${warehouse_password}" > ${warehouse_mysql_conf} && chmod 600 ${warehouse_mysql_conf}
+echo -e "[client]\npassword=${reporting_password}" > ${reporting_mysql_conf} && chmod 600 ${reporting_mysql_conf}
```
esotrope: secrets are copied to a tmp dir in order to execute the mysql scirpt. these are cleaned up by the script however if the script is prematurely aborted you could have secrets floating around for some duration on your machine where you don't expect them

malaffoon: Why not pass the password on the command-line? It's not great but it's not worse than writing it somewhere.

validation/validate-reporting-iab.sql
```diff
@@ -0,0 +1,162 @@
+USE reporting;
+
+CREATE TABLE IF NOT EXISTS iab_validation
```
malaffoon: Hmm. I don't want to make things harder but is there a way we can do this without requiring table creation. In production, i don't have a user account that has table creation permissions. Yes, i can create one and use that to do this but making less intrusive would be nice. I guess that would involve executing all these queries and concatenating them to a file (like you do with the ${report_query} but for all of these.

validation/validate-reporting-ica.sql
malaffoon: Since we're comparing data for internal consistency it doesn't matter which we do ... using `e.school_year` would be checking that that value was properly migrated for all exams. Checking `a.school_year` would only be checking that that value was properly migrated for assessments.

validation/README.md
malaffoon: seems good enough for now; i still like your idea of getting the secrets from another repository.

validation/validate-migration.sh
esotrope: The thinking here is, we don't know the environment this script will be executed in. if it's a multi-user machine the process with the password can be read in plain text by another user potentially.

validation/validate-reporting-iab.sql
esotrope: I definitely agree. There would be no awkward table hanging around too. It will take some additional orchestration but i'll see what i can do.

validation/validate-reporting-ica.sql
esotrope: Ah. should have clarified. what was possibly incorrect was `e.school_year` was being selected and the results were grouped by `a.school_year`

agorina:
Just curious: did you see any discrepancies between warehouse and reporting while testing this?

validation/validate-migration.sh
agorina: 👍

esotrope:
@agorina When running against dev:

```
possible issues detected:

total differences: 3334
total differences between warehouse and reporting ICAs:     1109
total differences between warehouse and reporting IABs:     1109
total differences between reporting and reporting OLAP ICAs:     1116
```

validation/sql/validate-warehouse-iab.sql
```diff
@@ -0,0 +1,163 @@
...
+    (SELECT max(testNum)
+     FROM iab_validation) + 1,
+    'total_iab';
+INSERT INTO iab_validation (testNum, result1)
+  SELECT
+    (SELECT max(testNum)
+     FROM iab_validation),
+    count(*)
+  FROM exam
+  WHERE type_id = 2;
```
agorina: Could you please add `WHERE deleted = 0` to here and anywhere else if a table has `deleted` column?

agorina:
Could you please address my comment about `deleted` and re-post the results?

validation/sql/validate-warehouse-iab.sql
esotrope: yes! thanks for checking that.

esotrope:
@agorina added `deleted=0` to the where clause for warehouse `asmt` and `exam`
I'm running into an unrelated issue with caching however. I think the remote post_validation table is either not getting cleared properly or mysql is caching query results. I see `asmt.natural_id` values show up in the results which i no longer reference in any sql file.

esotrope:
Overhauled the validation script. please re-review.
Main differences:
1. SQL scripts no longer rely on a table in the databases and are readonly
2. SQL scripts are put into their own files under their respective test name and schema name
3. The script loop prints out tests and results as they are performed
4. The script copies only failed (relevant) test results (csv/diff files) into the results directory

validation/validate-migration.sh
```diff
@@ -0,0 +1,221 @@
...
+# type|test_name|query,result,headers,csv
+declare -a tests=(
+    "ica|total-ica|total_exams"
+    "ica|total-ica-scores|total_scale_score,total_standard_error,total_performance_level"
+    "ica|total-ica-by-asmt-asmtyear-condition-complete|total_exams,assessment_id,assessment_year,administrative_condition,complete"
+    "ica|total-ica-by-school-district|total_exams,school_id,district_name,school_name"
+    "iab|total-iab|total_exams"
+    "iab|total-iab-scores|total_scale_score,total_standard_error,total_performance_level"
+    "iab|total-iab-by-asmt-asmtyear-condition-complete|total_exams,assessment_id,assessment_year,administrative_condition,complete"
+    "iab|total-iab-by-school-district|total_exams,school_id,district_name,school_name"
```
esotrope: i suppose i could have written this in a more powerful scripting language that actually has objects. oh well.

validation/validate-migration.sh
```diff
@@ -0,0 +1,221 @@
...
+    local elapsed_time="$(($end_time-$start_time))"
+    local elapsed_time_formatted=`date -u -r ${elapsed_time} +%T`
+
+    echo "completed in ${elapsed_time_formatted}"
+    echo ''
+}
+
+# entry point
+print_settings
+run_tests tests[@]
```
esotrope: nice part about this is you can tweak the script to run just the tests you want if running all of them is too slow.

validation/validate-migration.sh
```diff
@@ -0,0 +1,221 @@
...
+echo $''
+echo $' __   __                   __   __       ___    __                            __       ___  __   __  '
+echo $'|__) |  \ |  |     |\/| | / _` |__)  /\   |  | /  \ |\ |    \  /  /\  |    | |  \  /\   |  /  \ |__) '
+echo $'|  \ |__/ |/\|     |  | | \__> |  \ /~~\  |  | \__/ | \|     \/  /~~\ |___ | |__/ /~~\  |  \__/ |  \ '
+echo $''
+echo $''
+
+# process input
+
+source $1
```
malaffoon: What happens if i forget to pass in the secrets file or if it doesn't exist? Should it report a usage error?

esotrope: `source $1` will error out but I could add a more informative error message.

validation/validate-migration.sh
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env bash
```
malaffoon: fun script

validation/sql/warehouse/total-iab.sql
```diff
@@ -0,0 +1,5 @@
+SELECT
+    count(*)
+  FROM exam
+  WHERE type_id = 2
+    AND deleted = 0;
```
esotrope: @agorina in warehouse there is also `a.deleted` do i need to include that everywhere I include `e.deleted`?

validation/validate-migration.sh
esotrope: I thought so too : J

validation/sql/warehouse/total-iab.sql
agorina: You need to include `deleted` for every table you select from that has `deleted` column. It is enough to do it for one main table though, since if it is deleted there, any referenced tables should be cleaned up too. Let me know if this is confusing and we can chat offline.

It is because we soft-delete in the warehouse and hard-delete in the reporting.

validation/sql/warehouse/total-ica-by-asmt-asmtyear-condition-complete.sql
```diff
@@ -0,0 +1,20 @@
...
+    a.school_year,
+    ac.code,
+    CASE WHEN e.completeness_id = 2
+      THEN 'TRUE'
+    ELSE 'FALSE' END
+  FROM exam e
+    JOIN asmt a ON e.asmt_id = a.id
+    JOIN administration_condition ac ON e.administration_condition_id = ac.id
+  WHERE a.type_id = 1
+    AND a.deleted = 0
```
agorina: I do not think you need a.deleted here. 
If asmt is deleted we are not supposed to have exams for it.  And if we do - will have more problems. The migrate will fail in this case.

esotrope: cool - I'll go ahead and update these queries accordingly.

PR #191, opened 2017-11-21
Add configurable instructional resources.

wtritch:
There is a requirement that we can have 0-1 Instructional Resource link per assessment, organization level, performance level.

A null performance level indicates an assessment-wide instructional resource not bound to a single performance level.

IABs may have one instructional resource for each of the 3 "Reporting Catogories"
ICAs/Summatives may have one instructional resource for each of the 4 "Achievement Levels"

### DIFF
--- a/reporting/sql/V1_1_0_11__refactor_instructional_resource.sql
+++ b/reporting/sql/V1_1_0_11__refactor_instructional_resource.sql
@@ -1,0 +1,15 @@
+-- Refactor instructional_resource table to allow organization-level and performance-level
+-- instructional resources.
+
+USE ${schemaName};
+
+ALTER TABLE instructional_resource
+  DROP PRIMARY KEY,
+  CHANGE name asmt_natural_id VARCHAR(250) NOT NULL,
+  ADD COLUMN performance_level TINYINT DEFAULT NULL,
+  ADD COLUMN org_level VARCHAR(15) NOT NULL, -- 'SBAC' | 'State' | 'DistrictGroup' | 'District' | 'SchoolGroup'
+  ADD COLUMN org_natural_id VARCHAR(250) DEFAULT NULL,
+  ADD PRIMARY KEY (asmt_natural_id, org_level, performance_level);
+
+UPDATE instructional_resource
+  SET org_level = 'SBAC';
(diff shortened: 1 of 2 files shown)


reporting/sql/V1_1_0_11__refactor_instructional_resource.sql
```diff
@@ -0,0 +1,15 @@
+-- Refactor instructional_resource table to allow organization-level and performance-level
+-- instructional resources.
+
+USE ${schemaName};
+
+ALTER TABLE instructional_resource
+  DROP PRIMARY KEY,
+  CHANGE name asmt_natural_id VARCHAR(250) NOT NULL,
+  ADD COLUMN performance_level TINYINT DEFAULT NULL,
+  ADD COLUMN org_level VARCHAR(15) NOT NULL, -- 'SBAC' | 'State' | 'DistrictGroup' | 'District' | 'SchoolGroup'
```
esotrope: Should SBAC level just be called "System" level?

malaffoon: i agree that `SBAC` is not good, but i'm not sure what it exactly is. Is it meant to be the company that owns the data warehouse, the client configured in ART, or the assessment package vendor?

wtritch: I believe it's meant to be SBAC.  As in, this is their system for their assessments, and the reason they don't want it to be modifiable is expressly *because* they don't want anybody non-SBAC to fiddle with the links.

jtreuting: it is SBAC, but in a sense with an open source system it is the group that deploys and owns the system.  then they accept results for a tenant and levels below and they can customize.  so it should probably be a little more generic since it is not proprietary to SBAC (in theory at least)

wtritch: alrighty, in that case I'm leaning towards @esotrope's suggestion of System.

wtritch: done

PR #192, opened 2017-11-21
DWR-902 Add Bulk Delete Exams script

agorina:
There is also a documentation update in a separate PR.
Suggest to start from it: https://github.com/SmarterApp/RDW/pull/4

### DIFF
--- a/warehouse/sql/bulk_delete_exam.sql
+++ b/warehouse/sql/bulk_delete_exam.sql
@@ -4,7 +4,7 @@
 -- This may create a large volume of data to process. It is highly advisable to run this during the maintenance
 -- window and while the Exam Processor is stopped.
 --
--- Validation scripts should be executed before and after this scrip to check data quality:
+-- Validation script should be executed before and after this scrip to check data quality:
 -- https://github.com/SmarterApp/RDW_Schema/validation-scripts
 -- ------------------------------------------------------------------------------------------------------------
 
@@ -27,7 +27,7 @@ SET @partition_size = 1000000;
 -- create a unique digest fo the bulk run
 SELECT concat('bulk run ' COLLATE utf8_unicode_ci, now()) into @digest;
 
--- create a table to store and partition ids of the exams to be deleted
+-- create a table to store the partition ids of the exams to be deleted
 CREATE TABLE exam_delete_partition (
   exam_id BIGINT NOT NULL PRIMARY KEY,
   partition_id INT
@@ -48,11 +48,9 @@ CREATE PROCEDURE loop_by_partition(IN p_sql VARCHAR(1000), IN p_last INTEGER)
       SELECT concat('executed partition:', iteration);
 
       SET iteration = iteration + 1;
-      IF iteration <= p_last
-      THEN
-        ITERATE partition_loop;
+      IF iteration > p_last
+        THEN LEAVE partition_loop;
       END IF;
-      LEAVE partition_loop;
     END LOOP partition_loop;
 
   END;
@@ -70,8 +68,8 @@ DEALLOCATE PREPARE stmt;
 -- ------------------------------------------------------------------------------------------------------------
 -- STEP 5: partition exams to be process in chunks
 -- ------------------------------------------------------------------------------------------------------------
-SELECT floor(count(*)/@partition_size) INTO @exam_partition_end FROM exam_delete_partition;
-UPDATE exam_delete_partition SET partition_id = MOD(exam_id, @exam_partition_end + 1);
+SELECT ceil(count(*)/@partition_size) INTO @exam_partition_end FROM exam_delete_partition;
+UPDATE exam_delete_partition SET partition_id = MOD(exam_id, @exam_partition_end);
 
 -- ------------------------------------------------------------------------------------------------------------
 -- STEP 6: create imports - one per exam
@@ -101,7 +99,7 @@ SELECT max(id) INTO @maxImportId FROM import WHERE digest = @digest;
 
 -- distribute imports timestamps for migrate to have smaller chunks
 UPDATE import
-SET status = 1,
+SET
   created = DATE_ADD(created, INTERVAL (@maxImportId -id)  MICROSECOND),
   updated = DATE_ADD(updated, INTERVAL (@maxImportId -id)  MICROSECOND)
 WHERE status = 0
@@ -114,6 +112,13 @@ UPDATE exam e
 SET
   e.updated = i.updated
 WHERE i.status = 1
+      and content = 1
+      and digest = @digest;
+
+-- update import status to release it for the migrate
+UPDATE import
+  SET status = 1
+WHERE status = 0
       and content = 1
       and digest = @digest;
 
--- a/warehouse/sql/bulk_delete_exam.sql
+++ b/warehouse/sql/bulk_delete_exam.sql
@@ -111,7 +111,7 @@ UPDATE exam e
   JOIN import i ON i.id = e.update_import_id
 SET
   e.updated = i.updated
-WHERE i.status = 1
+WHERE i.status = 0
       and content = 1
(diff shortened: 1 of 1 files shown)


warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,129 @@
...
+UPDATE setting SET value = 'FALSE' WHERE name = 'AUDIT_TRIGGER_ENABLE';
+
+-- ------------------------------------------------------------------------------------------------------------
+-- STEP 3: initialization
+-- ------------------------------------------------------------------------------------------------------------
+-- define partition size
+SET @partition_size = 1000000;
+
+-- create a unique digest fo the bulk run
+SELECT concat('bulk run ' COLLATE utf8_unicode_ci, now()) into @digest;
```
agorina: I will research why this was needed.

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,129 @@
+-- ------------------------------------------------------------------------------------------------------------
+-- IMPORTANT NOTE
+-- ------------------------------------------------------------------------------------------------------------
+-- This may create a large volume of data to process. It is highly advisable to run this during the maintenance
+-- window and while the Exam Processor is stopped.
+--
+-- Validation scripts should be executed before and after this scrip to check data quality:
```
malaffoon: script

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,129 @@
...
+-- https://github.com/SmarterApp/RDW_Schema/validation-scripts
+-- ------------------------------------------------------------------------------------------------------------
+
+-- ------------------------------------------------------------------------------------------------------------
+-- STEP 1: replace a placeholder below
```
malaffoon: Please provide an example in the comment. Is it a clause, does it include `WHERE`, etc.

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,129 @@
...
+-- ------------------------------------------------------------------------------------------------------------
+-- STEP 3: initialization
+-- ------------------------------------------------------------------------------------------------------------
+-- define partition size
+SET @partition_size = 1000000;
+
+-- create a unique digest fo the bulk run
+SELECT concat('bulk run ' COLLATE utf8_unicode_ci, now()) into @digest;
+
+-- create a table to store and partition ids of the exams to be deleted
```
malaffoon: ... store *the* partition ...  ?

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,129 @@
...
+      EXECUTE stmt;
+      DEALLOCATE PREPARE stmt;
+      SELECT concat('executed partition:', iteration);
+
+      SET iteration = iteration + 1;
+      IF iteration <= p_last
+      THEN
+        ITERATE partition_loop;
+      END IF;
+      LEAVE partition_loop;
```
malaffoon: I don't know mysql LOOP (and i'm clearly too lazy to look it up) but it seems like there is extra stuff here: couldn't you just do:
```
    IF iteration > p_last THEN
      LEAVE partition_loop;
    END IF;
  END LOOP partition_loop;
```

agorina: And I do not know mysql LOOP (and clearly too lazy to look it up), I just copied it from the past version that was meant to be a one time run :)

Now I will have to check it out. Will do.

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,129 @@
...
+-- ------------------------------------------------------------------------------------------------------------
+SET @load_exam_delete_partion = concat('INSERT IGNORE INTO exam_delete_partition (exam_id) ', @select_exam_ids);
+PREPARE stmt FROM @load_exam_delete_partion;
+EXECUTE stmt;
+DEALLOCATE PREPARE stmt;
+
+-- ------------------------------------------------------------------------------------------------------------
+-- STEP 5: partition exams to be process in chunks
+-- ------------------------------------------------------------------------------------------------------------
+SELECT floor(count(*)/@partition_size) INTO @exam_partition_end FROM exam_delete_partition;
```
malaffoon: Nit. And i hesitate to even suggest it but ...

Instead of using floor() i'd be tempted to use ceil(). Then you wouldn't have the +1 in the MOD and the loop would be `< p_last` instead of `<= p_last`. Guess it depends if you're a 0-based or 1-based person.

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,129 @@
...
+-- distribute imports timestamps for migrate to have smaller chunks
+UPDATE import
+SET status = 1,
+  created = DATE_ADD(created, INTERVAL (@maxImportId -id)  MICROSECOND),
+  updated = DATE_ADD(updated, INTERVAL (@maxImportId -id)  MICROSECOND)
+WHERE status = 0
+      and content = 1
+      and digest =  @digest;
+
+-- update exams to match imports
```
malaffoon: i know i've copied this logic a couple times but isn't there a problem here? We're updating the import record status to PROCESSED *before* we've properly set the exam updated field. If migrate kicks in before the second update, what is going to happen?

agorina: Good point...but migrate does not care about the status. So if migrate is running all this time we have a problem. I need to add an instruction to pause migrate.( Unless you have a better idea).

warehouse/sql/bulk_delete_exam.sql
agorina: It will be in the markdown document.

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,134 @@
...
+-- STEP 3: initialization
+-- ------------------------------------------------------------------------------------------------------------
+-- define partition size
+SET @partition_size = 1000000;
+
+-- create a unique digest fo the bulk run
+SELECT concat('bulk run ' COLLATE utf8_unicode_ci, now()) into @digest;
+
+-- create a table to store the partition ids of the exams to be deleted
+CREATE TABLE exam_delete_partition (
```
jtreuting: should this be a temporary table?

agorina: I contemplated this idea but decided to make it permanent for 2 reasons:
1. nice to keep it for the troubleshooting purpose if something goes wrong
2. was not sure how stored procedure will like a temp table

If you feel strongly I can look into it more.

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -0,0 +1,136 @@
...
+  e.updated = i.updated
+WHERE i.status = 0
+      AND content = 1
+      AND digest = @digest;
+
+-- update import status to release it for the migrate
+UPDATE import
+SET
+  status = 1,
+  updated = updated -- make sure to not reset the timestamp
```
agorina: This was giving me a problem, fixed now.

I was able to wipe out report and restore it back using this script on my local machine.

warehouse/sql/bulk_delete_exam.sql
malaffoon: I thought about this last night and i think the issue that is bothering me is that we're modifying the collate for the value, not the encoding. If it works, i think this would be more correct:
```
SELECT concat(_utf8'bulk run ', now()) into @digest;
```

malaffoon: Hmm, i'm waffling on this. Perhaps it is actually the collation we want to specify since you're doing it so the comparison works. Meh. I'm probably overthinking this.

PR #193, opened 2017-11-21
Remove invalid instructional_resource primary key in favor of a uniqu…

wtritch:
…e index that allows nullables

While building out the API I found that the previous PRIMARY KEY on (asmt_natural_id, org_level, performance_level) doesn't work.  We may have an entry for those values for each organization at the org_level.

Unfortunately, this means the table cannot have a PRIMARY KEY because it would need to include the NULLABLE org_natural_id column.  To have a valid PRIMARY KEY we would need to introduce an auto-incrementing id column, but I didn't see much need.

Instead, I'm creating a UNIQUE INDEX on (asmt_natural_id, org_level, performance_level, org_natural_id) to ensure data integrity.

### DIFF
--- a/reporting/sql/V1_1_0_12__fix_instructional_resource_primary_key.sql
+++ b/reporting/sql/V1_1_0_12__fix_instructional_resource_primary_key.sql
@@ -4,4 +4,4 @@ USE ${schemaName};
 
 ALTER TABLE instructional_resource
   DROP PRIMARY KEY,
-  ADD UNIQUE INDEX idx__instructional_resource__asmt_level_perf_org (asmt_natural_id, org_level, performance_level, org_natural_id);
+  ADD UNIQUE INDEX idx__instructional_resource (asmt_natural_id, org_level, performance_level, org_natural_id);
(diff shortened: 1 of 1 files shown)


reporting/sql/V1_1_0_12__fix_instructional_resource_primary_key.sql
```diff
@@ -0,0 +1,7 @@
+-- Update instructional_resource primary key to allow a single value per institution
+
+USE ${schemaName};
+
+ALTER TABLE instructional_resource
+  DROP PRIMARY KEY,
+  ADD UNIQUE INDEX idx__instructional_resource__asmt_level_perf_org (asmt_natural_id, org_level, performance_level, org_natural_id);
```
wtritch: I had to munge the index name a bit "idx__instructional_resource__asmt_org_level_performance_level_org" was too long. =)

malaffoon:
Dang it, i need to trust my instincts ... i was going to ask about that in the last review.

reporting/sql/V1_1_0_12__fix_instructional_resource_primary_key.sql
malaffoon: IMO, for singular unique index and primary key it is sufficient to have just the table name, i.e. `pk__instructional_resource` or `idx__instructional_resource`. But i'm fine with the name you ended up with.

wtritch: Hm... that feels a lot cleaner.  Will rename to idx__instructional_resource

PR #194, opened 2017-11-22
improves readme run steps and simplifies script to use sql file disco…

esotrope:
…very instead of test_type configuration

### DIFF
--- a/validation/validate-migration.sh
+++ b/validation/validate-migration.sh
@@ -205,6 +205,10 @@ base_dir=`cd "$(dirname "$0")" ; pwd -P`
 sql_dir=${base_dir}/sql
 out_dir="${base_dir}/results-$(now_in_YYYY_mm_dd_HHMMSS)"
 diff_options="-y --suppress-common-lines"
+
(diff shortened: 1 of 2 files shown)


validation/validate-migration.sh
```diff
@@ -155,45 +189,25 @@ function print_summary() {
...
+validate_input
 
 # settings
-
 start_time=`now_in_seconds`
 base_dir=`cd "$(dirname "$0")" ; pwd -P`
 sql_dir=${base_dir}/sql
 out_dir="${base_dir}/results-$(now_in_YYYY_mm_dd_HHMMSS)"
 diff_options="-y --suppress-common-lines"
+warehouse_namespace=warehouse
```
malaffoon: nit: perhaps a comment that these are used both for the namespace in the database and some local directories so shouldn't be modified

esotrope: though they roughly relate to the database schema name they aren't used in the database. only in the naming of test sql files and output csv/diff files. Nevertheless, I'll add a comment 👍

PR #195, opened 2017-11-22
noticed this running locally

malaffoon:
If you look at the migrate scripts they round the scale_score when copying from warehouse to reporting. So i think the validation scripts need to do the same thing.

### DIFF
--- a/validation/sql/total-ica-scores/warehouse.sql
+++ b/validation/sql/total-ica-scores/warehouse.sql
@@ -1,5 +1,5 @@
 SELECT
-    sum(scale_score),
+    sum(round(scale_score)),
     sum(scale_score_std_err),
     sum(performance_level)
   FROM exam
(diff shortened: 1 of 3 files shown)


validation/sql/total-ica-scores/warehouse.sql
```diff
@@ -1,5 +1,5 @@
 SELECT
-    sum(scale_score),
+    sum(round(scale_score)),
```
esotrope: interesting. my only question is - will we round when migrating to the reporting olap schema?
If not there would need to be a different basis of comparison here.

malaffoon: Good point. We should keep that in mind.

warehouse/sql/bulk_delete_exam.sql
```diff
@@ -133,4 +134,4 @@ DROP PROCEDURE loop_by_partition;
 -- ------------------------------------------------------------------------------------------------------------
 -- STEP 10 (optional): turn auditing back on (if needed)
 -- ------------------------------------------------------------------------------------------------------------
-UPDATE setting SET value = 'TRUE' WHERE name = 'AUDIT_TRIGGER_ENABLE';
+UPDATE setting SET value = @audit_setting WHERE name = 'AUDIT_TRIGGER_ENABLE' AND value != @audit_setting;
```
agorina: Thank you!

chdietz: Oh, nice catch, put it back where it was.

PR #196, opened 2017-11-23
Add audit tables and triggers for student group related tables.

chdietz:
Add created timestamp to 
- student_group_membership 
- user_student_group

Add audit tables and triggers for 
- student_group 
- student_group_membership 
- user_student_group

Will push IT after this merge

### DIFF
--- a/warehouse/sql/V1_1_0_21__audit_groups.sql
+++ b/warehouse/sql/V1_1_0_21__audit_groups.sql
@@ -1,0 +1,159 @@
+-- Add audit tables and triggers for student groups
+
+USE ${schemaName};
+
+-- Add timestamp to student group child tables to support auditing
+-- Existing records need to have this value set in development, QA, staging and production.
+ALTER TABLE student_group_membership
+  ADD COLUMN created timestamp(6) default CURRENT_TIMESTAMP(6) not null;
+
+ALTER TABLE user_student_group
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_1_0_21__audit_groups.sql
```diff
@@ -0,0 +1,159 @@
+-- Add audit tables and triggers for student groups
+
+USE ${schemaName};
+
+-- Add timestamp to student group child tables to support auditing
+-- Existing records need to have this value set in development, QA, staging and production.
```
chdietz: Child tables need timestamp to query for auditing.

chdietz: Not sure about this what do we want to set value to for existing records?

malaffoon: Just pull the value from the student_group table.

chdietz: Since this is already merged this can be done in the merge of schema changes or I can add another flyway?

malaffoon: If i can somehow remember it during the merge, that will be fine. I'll add a note to my merge TODO

PR #198, opened 2017-11-27
adds embargo tables/columns

esotrope:
Adds warehouse & reporting schema changes.


reporting/sql/V1_1_0_14__embargo.sql
```diff
@@ -0,0 +1,10 @@
+USE ${schemaName};
+
+ALTER TABLE school
+  ADD embargoed tinyint NOT NULL;
```
agorina: Can a `school` or `school_group` be embargoed? Or is it `embargoed_district` flag?

warehouse/sql/V1_1_0_22__embargo.sql
```diff
@@ -0,0 +1,37 @@
...
+
+CREATE TABLE IF NOT EXISTS district_embargo (
+  district_id int NOT NULL,
+  embargo_scope_id int NOT NULL,
+  school_year smallint NOT NULL,
+  enabled tinyint NOT NULL,
+  deleted tinyint NOT NULL DEFAULT 0,
+  updated_on TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6) NOT NULL,
+  updated_by varchar(128),
+  PRIMARY KEY(district_id, embargo_scope_id, school_year, updated_on),
```
agorina: I do not think updated_on belongs here. Unless we are planning on using this table as a control AND audit one?

reporting/sql/V1_1_0_14__embargo.sql
esotrope: It is the district embargo flag making its way through affected organizations.
good to know about the nullable columns - thanks

warehouse/sql/V1_1_0_22__embargo.sql
esotrope: If we remove that do we still want the `updated_by` col?

reporting/sql/V1_1_0_14__embargo.sql
agorina: Could you please rename it to reflect that it is a district embargo?

warehouse/sql/V1_1_0_22__embargo.sql
agorina: Yes, it will just have the latest updated timestamp.

reporting/sql/V1_1_0_14__embargo.sql
esotrope: yes.
I think the olap schema would want a more generic column name though as the column will be a result of the state level and district level aggregate settings.

reporting/sql/V1_1_0_14__embargo.sql
```diff
@@ -0,0 +1,13 @@
+USE ${schemaName};
+
+ALTER TABLE school ADD district_embargo_enabled tinyint;
+UPDATE school SET district_embargo_enabled = 0;
+ALTER TABLE school MODIFY COLUMN district_embargo_enabled tinyint NOT NULL;
+
+ALTER TABLE school_group ADD district_embargo_enabled tinyint;
```
malaffoon: i'm sure i'll see it somewhere else but ... is this denormalization really necessary?

reporting/sql/V1_1_0_14__embargo.sql
```diff
@@ -0,0 +1,13 @@
+USE ${schemaName};
+
+ALTER TABLE school ADD district_embargo_enabled tinyint;
```
malaffoon: nit (and i realize personal preference): i prefer shorter field names where possible and i think `district_embargo` is sufficient in this case. Or `district_embargoed`.

warehouse/sql/V1_1_0_22__embargo.sql
```diff
@@ -0,0 +1,37 @@
...
+  (1, 'individual'),
+  (2, 'aggregate');
+
+CREATE TABLE IF NOT EXISTS district_embargo (
+  district_id int NOT NULL,
+  embargo_scope_id int NOT NULL,
+  school_year smallint NOT NULL,
+  enabled tinyint NOT NULL,
+  deleted tinyint NOT NULL DEFAULT 0,
+  updated_on TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6) NOT NULL,
```
malaffoon: We have used a different field naming scheme everywhere else: `updated`, `created`. I don't think we have an `updater` anywhere else but that would follow the `creator` convention. An alternative could be `modified` and `modifier`.

If we really want the two-word version i think this should be `updated_at`. But i prefer `updated`.

warehouse/sql/V1_1_0_22__embargo.sql
malaffoon: Why isn't the primary key just (district_id, embargo_scope_id, school_year)? It doesn't make sense to have more than one entry with those values the same.

warehouse/sql/V1_1_0_22__embargo.sql
```diff
@@ -0,0 +1,37 @@
...
+INSERT INTO embargo_scope (id, code) VALUES
+  (1, 'individual'),
+  (2, 'aggregate');
+
+CREATE TABLE IF NOT EXISTS district_embargo (
+  district_id int NOT NULL,
+  embargo_scope_id int NOT NULL,
+  school_year smallint NOT NULL,
+  enabled tinyint NOT NULL,
+  deleted tinyint NOT NULL DEFAULT 0,
```
malaffoon: I don't understand the `deleted` flag here. The embargo setting can be deleted? Isn't it just toggled true/false?

warehouse/sql/V1_1_0_22__embargo.sql
```diff
@@ -0,0 +1,37 @@
...
+  updated_by varchar(128),
+  PRIMARY KEY(district_id, embargo_scope_id, school_year, updated_on),
+  INDEX idx__district_embargo__district (district_id),
+  INDEX idx__district_embargo__embargo_scope (embargo_scope_id),
+  CONSTRAINT fk__district_embargo__district FOREIGN KEY (district_id) REFERENCES district(id) ON DELETE CASCADE,
+  CONSTRAINT fk__district_embargo__embargo_scope FOREIGN KEY (embargo_scope_id) REFERENCES embargo_scope(id)
+);
+
+CREATE TABLE IF NOT EXISTS state_embargo (
+  embargo_scope_id int NOT NULL,
```
malaffoon: I know it will be a single value right now but shouldn't there be a state field in there? Perhaps it could be state_abbrev and be a VARCHAR(2).

warehouse/sql/V1_1_0_22__embargo.sql
```diff
@@ -0,0 +1,37 @@
...
+  school_year smallint NOT NULL,
+  enabled tinyint NOT NULL,
+  deleted tinyint NOT NULL DEFAULT 0,
+  updated_on TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6) NOT NULL,
+  updated_by varchar(128),
+  PRIMARY KEY(embargo_scope_id, school_year, updated_on),
```
malaffoon: ditto about deleted, updated_* and primary key

warehouse/sql/V1_1_0_22__embargo.sql
```diff
@@ -0,0 +1,37 @@
...
+  (2, 'aggregate');
+
+CREATE TABLE IF NOT EXISTS district_embargo (
+  district_id int NOT NULL,
+  embargo_scope_id int NOT NULL,
+  school_year smallint NOT NULL,
+  enabled tinyint NOT NULL,
+  deleted tinyint NOT NULL DEFAULT 0,
+  updated_on TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6) NOT NULL,
+  updated_by varchar(128),
```
malaffoon: We have used varchar(255) for user login elsewhere.

warehouse/sql/V1_1_0_22__embargo.sql
```diff
@@ -0,0 +1,37 @@
+USE ${schemaName};
```
agorina: Just to note: I have used version 22, so you would need to rename this file.

warehouse/sql/V1_1_0_22__embargo.sql
esotrope: Will change it to `PRIMARY KEY(district_id, embargo_scope_id, school_year)`

reporting/sql/V1_1_0_14__embargo.sql
esotrope: It's not needed atm - will remove

warehouse/sql/V1_1_0_22__embargo.sql
esotrope: ill change to updated to follow the rest of the tables. 
Since we don't store `updated_by` in other tables (1) should we store it? (2) can it be naming convention?

warehouse/sql/V1_1_0_22__embargo.sql
esotrope: after talking with Alla I will be removing this - we won't be doing upsert in this table.

warehouse/sql/V1_1_0_22__embargo.sql
esotrope: that's a bit like telling the tenant what tenant it is - we store this in the app config right now

warehouse/sql/V1_1_0_22__embargo.sql
esotrope: 👍

warehouse/sql/V1_1_0_22__embargo.sql
esotrope: 👍

reporting/sql/V1_1_0_14__embargo.sql
esotrope: The idea here was to communicate that the value directly reflects `district_embargo.enabled`

esotrope:
I addressed all comments except the naming of `updated_by` and `district_embargo_enabled` and the addition of a `state_embargo.state_abbr` column.

reporting/sql/V1_1_0_14__embargo.sql
chdietz: Well the setting for audit is a value but the name is AUDIT_TRIGGER_ENABLE set to true or false.  I like district_embargo_enabled set to true or false, that way it is very clear. 

On a side note should we use ver or adjatiave, enable or enabled ?
A minor detail but it always nice to strive for consistency.

warehouse/sql/V1_1_0_23__embargo.sql
```diff
@@ -0,0 +1,35 @@
...
+  district_id int NOT NULL,
+  embargo_scope_id int NOT NULL,
+  school_year smallint NOT NULL,
+  enabled tinyint NOT NULL,
+  updated TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6) NOT NULL,
+  updated_by varchar(255),
+  PRIMARY KEY(district_id, embargo_scope_id, school_year),
+  INDEX idx__district_embargo__district (district_id),
+  INDEX idx__district_embargo__embargo_scope (embargo_scope_id),
+  CONSTRAINT fk__district_embargo__district FOREIGN KEY (district_id) REFERENCES district(id) ON DELETE CASCADE,
```
chdietz: ON DELETE CASCADE ? I don't see we have used this yet, is there a special need for this here?

warehouse/sql/V1_1_0_23__embargo.sql
```diff
@@ -0,0 +1,35 @@
...
+);
+
+INSERT INTO embargo_scope (id, code) VALUES
+  (1, 'individual'),
+  (2, 'aggregate');
+
+CREATE TABLE IF NOT EXISTS district_embargo (
+  district_id int NOT NULL,
+  embargo_scope_id int NOT NULL,
+  school_year smallint NOT NULL,
```
chdietz: What is the plan for transition to a new school year?  Will there be an operational process to populate this table with the new school year and districts?

esotrope: ah yes. great point. because we present historical exam data in reporting we will probably need a table to join on in reporting that has the embargo enabled+schoolYear combination. If we denormalized onto the exam itself we wouldn't need to do this because the exam has a specific `asmt.school_year`

warehouse/sql/V1_1_0_23__embargo.sql
```diff
@@ -0,0 +1,35 @@
...
+CREATE TABLE IF NOT EXISTS state_embargo (
+  embargo_scope_id int NOT NULL,
+  school_year smallint NOT NULL,
+  enabled tinyint NOT NULL,
+  updated TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6) NOT NULL,
+  updated_by varchar(255),
+  PRIMARY KEY(embargo_scope_id, school_year),
+  INDEX idx__state_embargo__embargo_scope (embargo_scope_id),
+  CONSTRAINT fk__state_embargo__embargo_scope FOREIGN KEY (embargo_scope_id) REFERENCES embargo_scope(id)
+);
```
chdietz: There is a state/tenant enable/disable embargo as a whole configuration.  Does this configuration feature need to be addressed in the schema?  Just wondering if a state does not want to use embargo the queries still need to work.

warehouse/sql/V1_1_0_23__embargo.sql
chdietz: Related to this I see there is a requirement to audit embargo changes, if so do we want to delete these records?

warehouse/sql/V1_1_0_23__embargo.sql
chdietz: What will trigger the migration for district_embargo and state_embargo?

malaffoon: Haven't started on that part yet. I'll be tackling it unless somebody gets to it first.

chdietz: ok, part of what I was thinking off is what master table will have the import id that drives the migration of embargo.

warehouse/sql/V1_1_0_23__embargo.sql
malaffoon: I think we should discuss the school-year aspect of embargo. I don't get the use-case of needing to embargo results from multiple years; it is always the current school year. OTOH, having the school-year in the embargo table *may* help. Guess we can fiddle with the queries and see where we end up.

chdietz: I was thinking that the cut over to a new school year will be safer and easier to manage with `school-year` column.

esotrope: That is my thinking as well.
Without a school year column we will have to rely on some scheduled task to flip the switches back at the turn of the school year.

reporting/sql/V1_1_0_14__embargo.sql
esotrope: If we add an embargo table to reporting i think it would look like this:
```sql
CREATE TABLE IF NOT EXISTS embargo (
  district_id int NOT NULL,
  enabled tinyint NOT NULL,
  PRIMARY KEY(district_id, school_year),
  INDEX idx__embargo__district (district_id),
  CONSTRAINT fk__embargo__district FOREIGN KEY (district_id) REFERENCES district(id)
);
```
Here would be an example of how we'd modify exam queries:
```sql
select e.*
from exam e
  join asmt a on e.asmt_id = a.id
  join school s on e.school_id = s.id
  left join embargo em on a.school_year = em.school_year and s.district_id = em.district_id
where not (e.type_id = 3 and em.enabled = 1)
```
@malaffoon @agorina @wtritch @chdietz thoughts? performance concerns?

agorina: I was under impression that in the `reporting` the embargo flag will be in the school table. Is this wrong?

esotrope: Wanted to do const/benefit analysis of this way. We can either:
1. store the flag in school/district table and have a reset switch that needs to be hit every year
2. join an embargo table and have no switch flipping

The big minus to the embargo table vs switch flip is the impact of an extra left join on all exam queries when most the exams are iab/ica.

agorina: I am afraid that it maybe pushing the boundaries of MySQL optimizer. The query will not be as simple as your example, there are also permissions clauses.  It is hard to guess without testing it with the large data set. Denormalizing it on the school table is definitely less concerning from the performance point of view. But I see the maintenance challenge with it. Ideally, if we have time, we can try it with this approach and load test against dev instance to see how it works. But if we do not have time and want to play safe, I would go with the school table solution.

chdietz: It's interesting to experiment with but in the data mart I tend to agree with Alla.

malaffoon:
@esotrope are we happy with this, can we commit it? Just need the warehouse side to start doing real work on the API.

malaffoon:
Working through the queries, i think there is an issue with district-groups: if we are going to allow the embargo admin permission to be set at the district-group level then i think we're going to have to add an optional district_group_id field to the district table and preserve that relationship during the organization import. Otherwise determining if a user has permissions for a district inherited through the district-group will be tricky. @esotrope, @wtritch thoughts?

esotrope:
@malaffoon - i think we'll need to support district group right? if that's the case we should probably migrate that relationship. Also I suppose this is ready to merge and we will figure out the reset later.

PR #199, opened 2017-11-28
Refactor instructional_resource table to reference assessments by nam…

wtritch:
…e rather than natural id

### DIFF
--- a/reporting/sql/V1_1_0_14__modify_instructional_resource_asmt.sql
+++ b/reporting/sql/V1_1_0_14__modify_instructional_resource_asmt.sql
@@ -7,3 +7,6 @@ ALTER TABLE instructional_resource
   DROP INDEX idx__instructional_resource,
   CHANGE asmt_natural_id asmt_name VARCHAR(250) NOT NULL,
   ADD UNIQUE INDEX idx__instructional_resource (asmt_name, org_level, performance_level, org_id);
+
+ALTER TABLE asmt
+  ADD INDEX idx__asmt__name (name);
--- a/reporting/sql/V1_1_0_14__modify_instructional_resource_asmt.sql
+++ b/reporting/sql/V1_1_0_14__modify_instructional_resource_asmt.sql
@@ -1,0 +1,9 @@
+-- Refactor instructional_resource table to reference asmt by name
+-- rather than natural_id.
+
+use ${schemaName};
+
+ALTER TABLE instructional_resource
+  DROP INDEX idx__instructional_resource,
+  CHANGE asmt_natural_id asmt_name VARCHAR(250) NOT NULL,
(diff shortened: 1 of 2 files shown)


reporting/sql/V1_1_0_14__modify_instructional_resource_asmt.sql
```diff
@@ -0,0 +1,9 @@
+-- Refactor instructional_resource table to reference asmt by name
+-- rather than natural_id.
+
+use ${schemaName};
+
+ALTER TABLE instructional_resource
+  DROP INDEX idx__instructional_resource,
+  CHANGE asmt_natural_id asmt_name VARCHAR(250) NOT NULL,
+  ADD UNIQUE INDEX idx__instructional_resource (asmt_name, org_level, performance_level, org_id);
```
wtritch: I debated adding an index on `asmt.name` but maybe the table is small enough that a string equality check will be performant enough?  Thoughts?

agorina: Yeah, I would add an index.

wtritch: done

malaffoon: Not suggesting a change but i'm surprised @agorina said that. For small tables doesn't MySQL typically optimize away the use of any index?

reporting/sql/V1_1_0_14__modify_instructional_resource_asmt.sql
```diff
@@ -0,0 +1,12 @@
+-- Refactor instructional_resource table to reference asmt by name
```
malaffoon: I'm curious why this change? Is it logically correct to use the name to key the resources? For two assessments (typically different school years), the resources will be the same?

wtritch: That's exactly the reason.  The goal is that we won't need to provide new performance-based or assessment-wide resources for the "same" assessment every year.

jtreuting: They created content for the ELA grade 11 brief write IAB, not the ELA grade 11 brief write IAB for 2015 and a different one for 2016.

reporting/sql/V1_1_0_14__modify_instructional_resource_asmt.sql
agorina: Just to explain: I admit I was lazy and did not check if MySQL optimizes it. My reason was that adding this index would not introduce any negative effect ( it is small in size and we do not update this table often), but **may** improve the query that will be executed often.

PR #200, opened 2017-11-28
DDWR-1132 Fix updated timestamp to reflect created/updated timestamp.

agorina:


warehouse/sql/V1_1_0_22__fix_accommodation_update.sql
```diff
@@ -0,0 +1,13 @@
+USE ${schemaName};
+
+UPDATE accommodation 
+ SET updated = now()
+WHERE updated is null;
```
agorina: It did not update the timestamp on create.
Another alternative is to find the latest import with content CODE and use a timestamp from there. 
Any preferences?

wtritch: I think setting it to "now" is completely reasonable.

warehouse/sql/V1_1_0_22__fix_accommodation_update.sql
```diff
@@ -0,0 +1,13 @@
...
+
+ALTER TABLE accommodation MODIFY COLUMN updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6);
```
malaffoon: foolish consistency should've caught this ... oops

agorina: true :)

PR #206, opened 2017-12-18
DWR-1242 Add views to support filling in missing data in the aggregate reports

agorina:


reporting_olap/sql/V1_0_0_3__add_views.sql
```diff
@@ -0,0 +1,50 @@
+-- add views to support filling in missing data in the aggregate reports
+
+SET SEARCH_PATH to ${schemaName};
+
+CREATE VIEW state_subject_grade_school_year(organization_id, organization_name, organization_type, subject_id, grade_id, school_year, asmt_id, asmt_type_id) AS
```
agorina: Note that all views have the same structure so that they could be used interchangeably in the final query.

reporting_olap/sql/V1_0_0_3__add_views.sql
```diff
@@ -0,0 +1,50 @@
...
+  SELECT
+    -1          AS id,
```
agorina: I need this magic number in order to support JOIN. Joining on NULLs does not work. It will never be exposed to the caller, it will be converted to NULL within the final query.

reporting_olap/sql/V1_0_0_3__add_views.sql
```diff
@@ -0,0 +1,50 @@
...
+    'state'     AS name,
+    'state'     AS organization_type,
+    s.id,
+    g.id,
+    year,
+    a.id as asmt_id,
+    a.type_id
+  FROM subject s
+    CROSS JOIN grade g
+      JOIN asmt a on a.grade_id = g.id and a.subject_id = s.id
```
agorina: Hmm...I think I could simplify it by removing a grade from here. Looking into it.

reporting_olap/sql/V1_0_0_3__add_views.sql
malaffoon: This should probably go in the file comments at the top

reporting_olap/sql/V1_0_0_3__add_views.sql
```diff
@@ -0,0 +1,50 @@
+-- add views to support filling in missing data in the aggregate reports
+
+SET SEARCH_PATH to ${schemaName};
+
+CREATE VIEW state_subject_grade_school_year(organization_id, organization_name, organization_type, subject_id, grade_id, school_year, asmt_id, asmt_type_id) AS
+  SELECT
+    -1          AS id,
+    'state'     AS name,
+    'state'     AS organization_type,
```
malaffoon: nit: we've been using `State` and `District` elsewhere

reporting_olap/sql/V1_0_0_3__add_views.sql
agorina: Done.

reporting_olap/sql/V1_0_0_3__add_views.sql
agorina: Changed.

PR #208, opened 2017-12-19
changes to embargo to support migrate

malaffoon:


reporting/sql/V1_1_0_16__migrate_embargo.sql
```diff
@@ -0,0 +1,24 @@
+-- Changes to embargo tables to support migrate
+
+USE ${schemaName};
+
+ALTER TABLE migrate
+  ADD COLUMN migrate_embargo tinyint
```
malaffoon: Like CODES, the EMBARGO migrate is "special" and doesn't have master-child tables.

reporting/sql/V1_1_0_16__migrate_embargo.sql
```diff
@@ -0,0 +1,24 @@
...
+ALTER TABLE district
+  DROP COLUMN embargo_enabled
+;
+
+CREATE TABLE IF NOT EXISTS staging_district_embargo (
+  district_id int NOT NULL,
+  individual tinyint,
+  aggregate tinyint,
+  migrate_id bigint NOT NULL
+);
```
malaffoon: We don't need `aggregate` for reporting, but i wanted to test the logic and have it there for later use (and/or for copying in olap migrate).

reporting/sql/V1_1_0_16__migrate_embargo.sql
chdietz: I guess if we get to a third case where we need a special flag in migrate for one entity type we should think about the design?

PR #209, opened 2017-12-21
DWR-1215 Merge scripts into one

agorina:
This is the first of the multiple steps to clean up olap migrate.
There are accompanying changes to both migrate and reporting projects.


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -34,25 +34,26 @@ CREATE TABLE staging_gender (
...
-  id int PRIMARY KEY NOT NULL,
+  id int PRIMARY KEY,
   grade_id smallint NOT NULL,
-  type_id smallint NOT NULL,
-  subject_id smallint NOT NULL,
   school_year smallint NOT NULL,
+  subject_id smallint NOT NULL,
+  type_id smallint NOT NULL,
   name character varying(250) NOT NULL,
+  label character varying(255) NOT NULL,
```
agorina: Added label, we need it for IABs.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -34,25 +34,26 @@ CREATE TABLE staging_gender (
...
   deleted boolean NOT NULL,
   migrate_id bigint NOT NULL,
   update_import_id bigint NOT NULL
- );
+) DISTSTYLE ALL;
 
 CREATE TABLE staging_district (
-  id int PRIMARY KEY NOT NULL,
+  id int NOT NULL PRIMARY KEY,
```
agorina: No change, just OCD.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -112,12 +113,19 @@ CREATE TABLE staging_exam_claim_score (
   migrate_id bigint NOT NULL
 );
 
+CREATE TABLE staging_school_year (
+  year smallint NOT NULL PRIMARY KEY SORTKEY
```
agorina: I added `NOT NULL` since it is the only constraint that Redshift enforces:http://docs.aws.amazon.com/redshift/latest/dg/t_Defining_constraints.html

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -148,26 +156,39 @@ CREATE TABLE administration_condition (
...
+  CONSTRAINT fk__asmt__grade FOREIGN KEY(grade_id) REFERENCES grade(id),
+  CONSTRAINT fk__asmt__school_year FOREIGN KEY(school_year) REFERENCES school_year(year)
+) DISTSTYLE ALL;
+
+CREATE TABLE active_asmt_per_year (
+  asmt_id int NOT NULL,
+  school_year smallint NOT NULL,
+  CONSTRAINT fk__active_asmt_per_yeart__asmt FOREIGN KEY(asmt_id) REFERENCES asmt(id),
+  CONSTRAINT fk__active_asmt_per_year__school_year FOREIGN KEY(school_year) REFERENCES school_year(year)
 ) DISTSTYLE ALL;
```
agorina: This table will be loaded by the migrate and is used to fill in the gap rows in the aggregate report.

jtreuting: guess i should have read this PR first, ignore my question from the other :)

reporting_olap/sql/V1_0_0_0__ddl.sql
malaffoon: i thought we didn't care about constraints for staging tables like this?

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -226,51 +247,73 @@ CREATE TABLE fact_student_ica_exam (
...
-  completed_at timestamptz encode lzo NOT NULL,
-  migrate_id bigint encode delta NOT NULL,
-  update_import_id bigint encode delta NOT NULL,
-  CONSTRAINT fk__fact_student_ica_exam__ica_asmt FOREIGN KEY(asmt_id) REFERENCES ica_asmt(id),
-  CONSTRAINT fk__fact_student_ica_exam__school FOREIGN KEY(school_id) REFERENCES school(id),
-  CONSTRAINT fk__fact_student_ica_exam__student FOREIGN KEY(student_id) REFERENCES student(id)
- )   COMPOUND SORTKEY (school_id, asmt_id);
-
+-- helper table used by the diagnostic API
 CREATE TABLE status_indicator (
```
malaffoon: nit (unrelated to these changes): this table name is a bit vague (both words in the name are generic); perhaps it could be renamed to indicate what it is used for or what might be found in it?

reporting_olap/sql/V1_0_0_0__ddl.sql
malaffoon: nit: it would be nice if this table name started with `asmt`, perhaps `asmt_active_year`

reporting_olap/sql/V1_0_0_0__ddl.sql
agorina: You are right about 'PRIMARY KEY SORTKEY', removing it.
Still want to keep NOT NULL, unless you have a reason to remove it.

reporting_olap/sql/V1_0_0_0__ddl.sql
agorina: `diagnostic_status`? 
Any other suggestions?

reporting_olap/sql/V1_0_0_0__ddl.sql
agorina: Will change the name.

PR #211, opened 2017-12-21
Add tables to support percentile

chdietz:
First stab at percentile tables to discuss with rdw_ingest and same branch name PR.

Updates:
renamed percentile flyway to lower number after combined schema changes.

percentile
 - name change: change number to count
  - make score a float as it is baed on scale_score that is a float
  - fix missing `on update` for `updated` timestamp

percentile_score
  - make compound primary key (any issue with float part of key?)
  - add contraint to parent table
  - keep calculated min and max for now but rename to min_score and max_score


warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,27 @@
+-- Add support for percentile
+
+USE ${schemaName};
+
+CREATE TABLE percentile (
+  id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  asmt_id INT NOT NULL,
+  number INT NOT NULL,
```
malaffoon: nit: not sure it's better but perhaps `count`?  `number` is just sooo generic

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,27 @@
...
+  mean FLOAT NOT NULL,
+  standard_deviation FLOAT NULL,
+  import_id BIGINT NOT NULL,
+  update_import_id BIGINT NOT NULL,
+  created TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) NOT NULL,
+  updated TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) NOT NULL,
```
malaffoon: `ON UPDATE CURRENT_TIMESTAMP(6)`?

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,27 @@
...
+);
+
+CREATE TABLE percentile_score (
+  id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  percentile_id INT NOT NULL,
+  percent TINYINT NOT NULL,
+  score INT NOT NULL,
+  min INT NOT NULL,
+  max INT NOT NULL
+);
```
malaffoon: `CONSTRAINT fk__percentile FOREIGN KEY (percentile_id) REFERENCES percentile (id)`?

Perhaps `PRIMARY KEY (percentile_id, percent)` instead of auto-increment id?

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,27 @@
...
+  updated TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) NOT NULL,
+  CONSTRAINT fk__percentile__asmt
+    FOREIGN KEY (asmt_id) REFERENCES asmt (id)
+);
+
+CREATE TABLE percentile_score (
+  id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  percentile_id INT NOT NULL,
+  percent TINYINT NOT NULL,
+  score INT NOT NULL,
```
malaffoon: What is `score`?  I guess i might see in the integration test code.

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,27 @@
...
+  min INT NOT NULL,
+  max INT NOT NULL
```
malaffoon: I never know what to call the values that define a percentile range. If the query is going to be like `min < value <= max` then it really shouldn't be `min`. 

Whatever they are called i think it should be more explicit like `min_score` and `max_score`.

warehouse/sql/V1_1_0_27__percentile.sql
jtreuting: looking at the integration test code it seems like it isn't really needed.  not sure it hurts to have it, but @chdietz is it needed in your mind?

warehouse/sql/V1_1_0_27__percentile.sql
chdietz: I can change that, got number from what Smarter used in their table

warehouse/sql/V1_1_0_27__percentile.sql
chdietz: good catch, wondering how I did that it looks like the `ON UPDATE CURRENT_TIMESTAMP(6)` is left off when you generate ddl from an existing table in intellij.

warehouse/sql/V1_1_0_27__percentile.sql
chdietz: Did not get to annotate this commit yet.  I was thinking that score is what is provided and min and max is what we calculate.  So I wanted to preserve what they give us in the warehouse but in reporting I think we do not need it.

warehouse/sql/V1_1_0_27__percentile.sql
chdietz: sounds good

warehouse/sql/V1_1_0_27__percentile.sql
chdietz: yes, that was clearly an oversight

warehouse/sql/V1_1_0_27__percentile.sql
chdietz: Another thought we talked about since score, min and max represent possible scale_score and scale_score is a float then these should be float also. Had it that way at first and changed to an int.
When we migrate to reporting maybe round and store as int?

malaffoon: I think the warehouse table should store just the `score` as a float.
When we migrate to reporting we can denormalize to `minScore`, `maxScore` as ints.

chdietz: I think that will work.

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,25 @@
+-- Add support for percentile
+
+USE ${schemaName};
+
+CREATE TABLE percentile (
+  id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  asmt_id INT NOT NULL,
+  count INT NOT NULL,
```
chdietz: changed number to count

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,25 @@
...
+  update_import_id BIGINT NOT NULL,
+  created TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+  CONSTRAINT fk__percentile__asmt FOREIGN KEY (asmt_id) REFERENCES asmt (id)
+);
+
+CREATE TABLE percentile_score (
+  percentile_id INT NOT NULL,
+  percent TINYINT NOT NULL,
+  score FLOAT NOT NULL,
```
chdietz: make score a float as it is the `scale_score` for a percentile.
- removed min and max as they are calculated from the provided percentile table.
- they will be name something like min_score and max_score

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,25 @@
...
+CREATE TABLE percentile (
+  id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  asmt_id INT NOT NULL,
+  count INT NOT NULL,
+  mean FLOAT NOT NULL,
+  standard_deviation FLOAT NULL,
+  import_id BIGINT NOT NULL,
+  update_import_id BIGINT NOT NULL,
+  created TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
```
chdietz: `ON UPDATE` was missing

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,25 @@
...
+  CONSTRAINT fk__percentile__asmt FOREIGN KEY (asmt_id) REFERENCES asmt (id)
+);
+
+CREATE TABLE percentile_score (
+  percentile_id INT NOT NULL,
+  percent TINYINT NOT NULL,
+  score FLOAT NOT NULL,
+  PRIMARY KEY (percentile_id, percent),
+  CONSTRAINT fk__percentile_score__percentile FOREIGN KEY (percentile_id) REFERENCES percentile (id)
```
chdietz: use compound key instead of unique id column

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,27 @@
...
+  updated TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6),
+  CONSTRAINT fk__percentile__asmt FOREIGN KEY (asmt_id) REFERENCES asmt (id)
+);
+
+CREATE TABLE percentile_score (
+  percentile_id INT NOT NULL,
+  percent TINYINT NOT NULL,
+  score FLOAT NOT NULL,
+  min_score SMALLINT NOT NULL,
+  max_score SMALLINT NOT NULL,
```
chdietz: Decided to keep min_score and max_score in warehouse to calculate on ingest as part of validation

warehouse/sql/V1_1_0_27__percentile.sql
```diff
@@ -0,0 +1,27 @@
+-- Add support for percentile
```
malaffoon: File name should probably be changed, e.g. V1_1_0_2__percentile.sql

chdietz: Thanks i was going to ask you that

chdietz:
@malaffoon @jtreuting I think comments have been addressed anything else before I merge?

PR #212, opened 2017-12-21
merge

malaffoon:
I know this is a hard-to-read PR since everything looks like its been deleted and added new. And there is some merging of scripts. But please, please, please carefully review any area of the schema that you are familiar with and make sure i didn't bork anything. 

I will let this sit for a little while because i need to figure out how to benchmark the execution against a large database (without actually ruining said large database).


reporting/sql/V1_0_2_0__remove_wer_warning.sql
```diff
@@ -1,14 +0,0 @@
--- Update the system-news label to remove the WER warning
```
malaffoon: This was thrown away since the change was manually applied in production.

reporting/sql/V1_1_0_0__update.sql
```diff
@@ -0,0 +1,172 @@
...
+  ADD COLUMN type varchar(40),
+  ADD COLUMN options_count tinyint,
+  ADD COLUMN answer_key varchar(50);
+
+
+ALTER TABLE instructional_resource
+  DROP PRIMARY KEY,
+  CHANGE name asmt_name VARCHAR(250) NOT NULL,
+  ADD COLUMN org_level VARCHAR(15) NOT NULL DEFAULT 'System',
+  ADD COLUMN performance_level TINYINT DEFAULT 0,
```
malaffoon: I twigged this a bit to use default values that reflect the current resources (all of them are System for any performance level). This was in preference to adding column without constraints, setting values, then turning back on constraints. Thoughts? Am i being a bit too lazy here?

warehouse/sql/V1_1_0_0__update.sql
```diff
@@ -0,0 +1,440 @@
...
+-- When first created, RDW_Schema was on build #273 and this incorporated:
+--   V1_0_0_2__add_index_to_import.sql
+--   V1_1_0_1__add_school_groups.sql
+--   ...
+--   V1_1_0_26__embargo_cleanup.sql
+
+USE ${schemaName};
+
+-- TODO - there is already a (updated, status) index; can one of these be removed?
+ALTER TABLE import ADD INDEX idx__import__status_updated (status, updated);
```
malaffoon: @agorina any thoughts on this?

warehouse/sql/V1_1_0_0__update.sql
```diff
@@ -0,0 +1,440 @@
...
+  END;
+//
+DELIMITER ;
+
+
+-- exam -----------------------------------------------------------------------------------------
+
+-- merge exam_student table into exam table -----------------------------------------------------
+
+ALTER TABLE exam
```
malaffoon: Modifying tables is expensive so i'm trying to consolidate the exam table changes as much as possible.

warehouse/sql/V1_1_0_0__update.sql
```diff
@@ -0,0 +1,440 @@
...
+  ADD COLUMN theta_score float,
+  ADD COLUMN theta_score_std_err float,
+  -- modify completed_at to have the same precision as other timestamps
+  MODIFY COLUMN completed_at TIMESTAMP(6) NOT NULL,
+  -- modify size of the t3_program_type column to account for all valid values
+  MODIFY COLUMN t3_program_type varchar(30);
+
+-- add and set partition
+-- TODO - how many partitions for production?
+SET @exam_partitions = 50;
```
malaffoon: @agorina thoughts?

warehouse/sql/V1_1_0_0__update.sql
```diff
@@ -0,0 +1,440 @@
...
+
+-- TODO - would it be faster to add index *after* setting all the values? enough faster?
```
malaffoon: Thoughts?

reporting/sql/V1_1_0_0__update.sql
jtreuting: that doesn't feel wrong to me

jtreuting:
Is the new added file supposed to cover everything from the delete files?  I see a lot of deleted files that create the triggers but nothing in the new file to account for that?

malaffoon:
There are two new files and, yes, everything should be covered. A lot of the triggers in the deleted files are updates so, when i merged them, only the last one counted.

jtreuting:
ahh i see it now.  it was hidden by default and I missed it.  then searched for the trigger and didn't see it anywhere but the delete files cause that 2nd new file wasn't expanded.

warehouse/sql/V1_1_0_0__update.sql
```diff
@@ -0,0 +1,450 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS state_embargo (
+  school_year smallint NOT NULL,
+  individual tinyint,
+  aggregate tinyint,
+  updated TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6) NOT NULL,
+  updated_by varchar(255),
+  PRIMARY KEY(school_year)
+);
```
esotrope: 👍 👍

warehouse/sql/V1_1_0_0__update.sql
agorina: I would not remove it without testing. I believe I made an attmept to remove one of them at some point, but found that both are needed. Things may have changed since then though.

 `idx__import__updated_status` was added here to help with the migration: https://github.com/SmarterApp/RDW_Schema/pull/116/files 

`idx__import__status_updated`  was introduced here to helo with the monitoring queries: 
https://github.com/SmarterApp/RDW_Schema/pull/155/files

warehouse/sql/V1_1_0_0__update.sql
agorina: Was this number used while testing on AWS dev (and 25 mil records) ?

If I recall it correctly we used about 500,000 per each partition during the initial load.
I think that up to 1 mil is a good size. 

I would do something like this:
```sql
SELECT ceil(count(*)/1000000) INTO @exam_partitions FROM exam;
```

malaffoon: Good idea, i'll change to use that.

PR #214, opened 2017-12-27
DWR-1215 Migrate district/school groups (olap)

agorina:


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -49,18 +49,40 @@ CREATE TABLE staging_asmt (
 CREATE TABLE staging_district (
   id int NOT NULL PRIMARY KEY,
   name character varying(100) NOT NULL,
+  natural_id varchar(40) NOT NULL,
+  external_id varchar(40),
```
agorina: I am including these too since I could see this being used in the export.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -156,17 +178,43 @@ CREATE TABLE administration_condition (
   code character varying(20) NOT NULL UNIQUE
 ) DISTSTYLE ALL;
 
-CREATE TABLE  district (
+CREATE TABLE district_group (
+  id integer encode raw NOT NULL PRIMARY KEY SORTKEY,
```
malaffoon: Why do we think we'll need these?

Obviously, we need the schools' membership in groups for permissions but that can be in the school table. Is it for that and this is just more efficient or is there another need?

agorina: I am not sure if this is needed but decided to include just in case. 
Open to remove if you think it is not needed.

PR #215, opened 2017-12-27
DWR-1215 Migrate embargo

agorina:
@chdietz, I know that the file name conflicts with your PR. One of us will have to change the name (depending on who merges it first)


warehouse/sql/V1_1_0_2__embargo.sql
```diff
@@ -0,0 +1,26 @@
...
+*/
+CREATE VIEW embargo AS
+  SELECT
+    d.id AS district_id,
+    de.school_year,
+    IF(se.individual = 0 OR de.individual = 0 OR (se.individual IS NULL AND de.individual IS NULL), 0, 1) AS individual,
+    IF(se.aggregate = 0 OR de.aggregate = 0, 0, 1) AS aggregate
+  FROM district d
+    LEFT JOIN district_embargo de ON de.district_id = d.id
+    LEFT JOIN state_embargo se ON se.school_year = de.school_year;
```
agorina: I copied it from the reporting migrate and converted into a view to reuse it in the olap migrate.

warehouse/sql/V1_1_0_2__embargo.sql
```diff
@@ -0,0 +1,26 @@
...
+embargo is FALSE and "default" for aggregate embargo is TRUE:
+
+             (individual)  (aggregate)
+    district  T   F  null   T   F  null
+             -----------   -----------
+ state T    | T | F | T |  | T | F | T |
+       F    | F | F | F |  | F | F | F |
+       null | T | F | F |  | T | F | T |
+             -----------    -----------
+ TODO - it would be slightly more efficient to get just districts that have schools associated with them
```
agorina: @malaffoon , it this TODO applicable here? ( I am not sure what it means)

warehouse/sql/V1_1_0_2__embargo.sql
```diff
@@ -0,0 +1,26 @@
+USE ${schemaName};
+
+/*
+Create records with effective embargo for ALL districts regardless of timestamps. The
+truth table for how state and district setting resolve to the effective setting is a bit
+non-obvious. Recall that state=FALSE trumps district setting, "default" for individual
+embargo is FALSE and "default" for aggregate embargo is TRUE:
+
+             (individual)  (aggregate)
```
agorina: @malaffoon , I added this line - does it look right?

warehouse/sql/V1_1_0_2__embargo.sql
chdietz: only diff I see is this is build for all school year and migrate sql had school year as a param.

chdietz: ok, so in migrate sql you select a specific school year from the view

agorina: That is correct. I do not know the year until the run time.

warehouse/sql/V1_1_0_2__embargo.sql
malaffoon: In production, there are a bunch of districts that have no schools associated with them. However, that is cruft that is left over from bad data. I think you can/should remove the TODO comment since the system shouldn't have to worry about districts with no schools associated with them.

warehouse/sql/V1_1_0_2__embargo.sql
malaffoon: yes, thanks

warehouse/sql/V1_1_0_2__embargo.sql
agorina: Ok, TODO is gone.

PR #216, opened 2017-12-28
DWR-1215 Add embargo to Redshift

agorina:
Not sure if this is right, but unlike reporting I think we may need the embargo flag on all three levels: school, district and state. Please let me know if this is incorrect.

### DIFF
--- a/reporting_olap/sql/V1_0_0_0__ddl.sql
+++ b/reporting_olap/sql/V1_0_0_0__ddl.sql
@@ -85,12 +85,12 @@ CREATE TABLE staging_school_group (
 
 CREATE TABLE staging_district_embargo (
   district_id integer NOT NULL,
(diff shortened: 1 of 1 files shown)


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -83,6 +83,17 @@ CREATE TABLE staging_school_group (
   migrate_id bigint NOT NULL
 );
 
+CREATE TABLE staging_district_embargo (
+  district_id integer NOT NULL,
+  aggregate smallint,
```
agorina: I did not see a need for the `individual` here. Please correct me if this is wrong.

agorina:
I removed state and district level flags. If needed, will add them later.

PR #217, opened 2017-12-29
DWR-1215 Clean up Redshift tables

agorina:
I would like to get people's feedback on the data elements I am about to remove.

I believe this is the final step for this JIRA. Please let me know if you see any data elements missing.

### DIFF
--- a/reporting_olap/sql/V1_0_0_0__ddl.sql
+++ b/reporting_olap/sql/V1_0_0_0__ddl.sql
@@ -42,7 +42,7 @@ CREATE TABLE staging_asmt (
   label character varying(255) NOT NULL,
   deleted boolean NOT NULL,
   migrate_id bigint NOT NULL,
-  updated timestamptz NOT NULL,
+  updated timestamp without time zone NOT NULL,
   update_import_id bigint NOT NULL
 ) DISTSTYLE ALL;
 
@@ -64,7 +64,7 @@ CREATE TABLE staging_school (
   district_group_id integer,
   deleted boolean NOT NULL,
   migrate_id bigint NOT NULL,
-  updated timestamptz NOT NULL,
+  updated timestamp without time zone NOT NULL,
   update_import_id bigint NOT NULL
 );
 
@@ -100,7 +100,7 @@ CREATE TABLE staging_student (
   gender_id smallint,
   deleted boolean NOT NULL,
   migrate_id bigint NOT NULL,
-  updated timestamptz NOT NULL,
+  updated timestamp without time zone NOT NULL,
   update_import_id bigint NOT NULL
  );
 
@@ -131,7 +131,7 @@ CREATE TABLE staging_exam (
   deleted boolean NOT NULL,
   completed_at timestamp without time zone NOT NULL,
   migrate_id bigint NOT NULL,
-  updated timestamptz NOT NULL,
+  updated timestamp without time zone NOT NULL,
   update_import_id bigint NOT NULL,
   latest boolean
 );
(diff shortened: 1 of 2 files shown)


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -9,8 +9,7 @@ SET client_encoding = 'UTF8';
 -- staging tables
 CREATE TABLE staging_grade (
   id smallint NOT NULL PRIMARY KEY,
-  code character varying(2) NOT NULL UNIQUE,
-  name character varying(100) NOT NULL UNIQUE
```
agorina: @esotrope , @wtritch : Would we ever need a name here?

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -96,13 +97,10 @@ CREATE TABLE staging_state_embargo (
 
 CREATE TABLE staging_student (
   id int NOT NULL PRIMARY KEY,
-  ssid character varying(65) NOT NULL,
-  last_or_surname character varying(60),
-  first_name character varying(60),
-  middle_name character varying(60),
```
agorina: I do not think we would ever need it in the aggregate reports.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -169,14 +168,12 @@ CREATE TABLE subject (
...
   id smallint NOT NULL PRIMARY KEY SORTKEY,
-  code character varying(2) NOT NULL UNIQUE,
-  name character varying(100) NOT NULL UNIQUE
+  code character varying(2) NOT NULL UNIQUE
 ) DISTSTYLE ALL;
 
 CREATE TABLE asmt_type (
   id smallint NOT NULL PRIMARY KEY SORTKEY,
-  code character varying(10) NOT NULL UNIQUE,
-  name character varying(24) NOT NULL UNIQUE
```
agorina: @wtritch , @esotrope : Ditto about the name?

reporting_olap/sql/V1_0_0_0__ddl.sql
jtreuting: agreed

reporting_olap/sql/V1_0_0_0__ddl.sql
wtritch: Nope, I think we exclusively use codes and rely on translation for naming

reporting_olap/sql/V1_0_0_0__ddl.sql
chdietz: saves space and also removes some PII where we do not need it

reporting_olap/sql/V1_0_0_0__ddl.sql
chdietz: So the 'grade' displayed is translated from the code, probably nice to have name in warehouse, but do we need `grade.name` in reporting? Should it be removed there also?

agorina: There is more than one field in reporting that could be cleaned up. 
Not sure if this is a priority.
 I will create a JIRA to review and remove unused fields from reporting.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -43,6 +42,7 @@ CREATE TABLE staging_asmt (
   label character varying(255) NOT NULL,
   deleted boolean NOT NULL,
   migrate_id bigint NOT NULL,
+  updated timestamp without time zone NOT NULL,
```
agorina: @wtritch, I copied this from 'completed_at' implementation. 
Do you remember why you choose to do it this way? 
I admit that I have played with different options and none seemed to work.

@malaffoon , just FYI - so that you see this change and the response.

wtritch: It had to do with preserving the non-time-zoned Instant from the TRT through import, migration, and the Reporting API to prevent shifting the Instant based upon server/DB time zones

PR #218, opened 2017-12-29
Feature/percentile reporting

chdietz:
Tables for reporting and staging

### DIFF
--- a/reporting/sql/V1_1_0_1__percentile.sql
+++ b/reporting/sql/V1_1_0_1__percentile.sql
@@ -1,0 +1,46 @@
+-- Add support for percentile
+
+USE ${schemaName};
+
+CREATE TABLE percentile (
+  id INT NOT NULL PRIMARY KEY,
+  asmt_id INT NOT NULL,
+  count INT NOT NULL,
+  mean FLOAT NOT NULL,
+  standard_deviation FLOAT NULL,
+  update_import_id BIGINT NOT NULL,
+  updated TIMESTAMP(6) NOT NULL,
+  migrate_id BIGINT NOT NULL,
+  CONSTRAINT fk__percentile__asmt FOREIGN KEY (asmt_id) REFERENCES asmt (id)
+);
+
+CREATE TABLE percentile_score (
+  percentile_id INT NOT NULL,
+  percent TINYINT NOT NULL,
+  score FLOAT NOT NULL,
+  min_score SMALLINT NOT NULL,
+  max_score SMALLINT NOT NULL,
+  PRIMARY KEY (percentile_id, percent),
+  CONSTRAINT fk__percentile_score__percentile FOREIGN KEY (percentile_id) REFERENCES percentile (id)
+);
+
+CREATE TABLE staging_percentile (
+  id INT NOT NULL PRIMARY KEY,
+  asmt_id INT NOT NULL,
+  count INT NOT NULL,
+  mean FLOAT NOT NULL,
+  standard_deviation FLOAT NULL,
+  update_import_id BIGINT NOT NULL,
+  updated TIMESTAMP(6) NOT NULL,
+  migrate_id BIGINT NOT NULL,
+  deleted TINYINT NOT NULL
+);
+
+CREATE TABLE staging_percentile_score (
+  percentile_id INT NOT NULL,
+  percent TINYINT NOT NULL,
+  score FLOAT NOT NULL,
+  min_score SMALLINT NOT NULL,
+  max_score SMALLINT NOT NULL,
+  PRIMARY KEY (percentile_id, percent)
+);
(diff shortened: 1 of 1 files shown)


reporting/sql/V1_1_0_1__percentile.sql
```diff
@@ -0,0 +1,46 @@
...
+CREATE TABLE staging_percentile (
+  id INT NOT NULL PRIMARY KEY,
+  asmt_id INT NOT NULL,
+  count INT NOT NULL,
+  mean FLOAT NOT NULL,
+  standard_deviation FLOAT NULL,
+  update_import_id BIGINT NOT NULL,
+  updated TIMESTAMP(6) NOT NULL,
+  migrate_id BIGINT NOT NULL,
+  deleted TINYINT NOT NULL
```
chdietz: Left off deleted in warehouse will have to add it.  We should have support to delete a percentile table and it should delete corresponding percentile_score(s)

reporting/sql/V1_1_0_1__percentile.sql
```diff
@@ -0,0 +1,46 @@
...
+  update_import_id BIGINT NOT NULL,
+  updated TIMESTAMP(6) NOT NULL,
+  migrate_id BIGINT NOT NULL,
+  CONSTRAINT fk__percentile__asmt FOREIGN KEY (asmt_id) REFERENCES asmt (id)
+);
+
+CREATE TABLE percentile_score (
+  percentile_id INT NOT NULL,
+  percent TINYINT NOT NULL,
+  score FLOAT NOT NULL,
```
chdietz: We may not need this column but I do not see the space gain in not having this column based on the number of expected rows.

reporting/sql/V1_1_0_1__percentile.sql
```diff
@@ -0,0 +1,46 @@
...
+  min_score SMALLINT NOT NULL,
+  max_score SMALLINT NOT NULL,
+  PRIMARY KEY (percentile_id, percent),
```
chdietz: I notice other tables such as `exam_available_accommodation` have indexes such as:
`UNIQUE INDEX idx__exam_available_accommodation (exam_id, accommodation_id)`

Is this not effectively the same as `PRIMARY KEY`?

reporting/sql/V1_1_0_1__percentile.sql
```diff
@@ -0,0 +1,46 @@
+-- Add support for percentile
+
+USE ${schemaName};
+
+CREATE TABLE percentile (
+  id INT NOT NULL PRIMARY KEY,
+  asmt_id INT NOT NULL,
+  count INT NOT NULL,
+  mean FLOAT NOT NULL,
+  standard_deviation FLOAT NULL,
```
malaffoon: i have no real reason to think that this isn't going to be the std.deviation but did we confirm that they are providing the std.deviation and not the std error?

reporting/sql/V1_1_0_1__percentile.sql
malaffoon: what about the confusion factor?

reporting/sql/V1_1_0_1__percentile.sql
malaffoon: The way MySQL works we needed to take control of the indexes. @agorina probably remembers the details. For a table this size the index doesn't matter much.

reporting/sql/V1_1_0_1__percentile.sql
malaffoon: I'm not entirely sure i agree with this (for example, our codes tables don't have deleted flags; and, even though the asmt table has a deleted flag, i'm pretty sure it isn't used and won't be with this release) but i guess it can't hurt. 

I guess an argument could be made that the RDW might one day be the management repository of the assessment norms -- when that happens it would be more likely that we want audit capability for these tables.

reporting/sql/V1_1_0_1__percentile.sql
agorina: The only issue I remember was about FK: https://github.com/SmarterApp/RDW_Schema/pull/94#pullrequestreview-45194970

reporting/sql/V1_1_0_1__percentile.sql
chdietz: I was thinking about the life cycle of these tables and that we display it if it exists.  So if a table is ingested, then a new version of the same table was re generated for whatever reason we would update(replace) it.  

But if it is determined the ingested table is not valid for some reason and there is no suitable replacement and they want it removed we could soft delete it in the warehouse and migrate would delete from reporting.

malaffoon: And that's fine. The alternative is we treat it like the codes tables: always replace the contents in the reporting data mart to make it match what's in the data warehouse.

reporting/sql/V1_1_0_1__percentile.sql
chdietz: ok, that was the issue of dropping a FK does not drop the underlying index, sort of a different issue.

reporting/sql/V1_1_0_1__percentile.sql
chdietz: @malaffoon I double checked, we asked, offered: Number, mean and standard deviation, and got no reply.

PR #223, opened 2018-01-08
DWR-1217 Fixing psql run.

agorina:


validation/validate-migration.sh
```diff
@@ -57,22 +57,12 @@ function create_mysql_password_file() {
...
-    rm ${password_file}
+    mysql -h ${connection[0]} -P ${connection[1]} -u ${connection[3]} ${connection[2]} -p${connection[4]} -s < ${sql_file} | tr '\t' ','
 }
 
 function psql_to_csv() {
     declare -a connection=("${!1}")
-    local sql_file=$2
-    local csv_file=$3
-    local csv_headers=$4
-    set PGPASSWORD=${connection[5]}
```
agorina: This was a typo, password is `${connection[4]}`.

But, per @malaffoon suggestion I am changing it to a one line call to avoid leaving behind a file with the password in case something fails.

validation/validate-migration.sh
```diff
@@ -57,22 +57,12 @@ function create_mysql_password_file() {
 
 function mysql_to_csv() {
     declare -a connection=("${!1}")
-    local sql_file=$2
-    local csv_file=$3
-    local csv_headers=$4
-    local password_file=`create_mysql_password_file "${connection[4]}"`
-    mysql --defaults-extra-file=${password_file} -h ${connection[0]} -P ${connection[1]} -u ${connection[3]} ${connection[2]} -s < ${sql_file} | tr '\t' ','
-    rm ${password_file}
+    mysql -h ${connection[0]} -P ${connection[1]} -u ${connection[3]} ${connection[2]} -p${connection[4]} -s < ${sql_file} | tr '\t' ','
```
agorina: This now gives a warning. Bit this does remove a risk of leaving a password file behind in a case of failure.

esotrope: We should add to the readme that the connection passwords will be visible when running the `ps` command then

PR #225, opened 2018-01-10
Bug/dwr 1294

agorina:


warehouse/sql/V1_1_0_5__drop_embargo.sql
```diff
@@ -0,0 +1,3 @@
+USE ${schemaName};
+
+DROP VIEW embargo;
```
agorina: I moved the code into a common sql.yml file in the application.

PR #226, opened 2018-01-10
DWR-1296 Add item performance task writing type

agorina:


reporting/sql/V1_1_0_2__add_item_pt_writing_type.sql
```diff
@@ -0,0 +1,4 @@
+USE ${schemaName};
+
+ALTER TABLE staging_item ADD COLUMN performance_task_writing_type varchar(16);
```
agorina: The longest value is 'Argumentative'. I just padded it a little. If people do not like the number 16 please suggest the one you like :)

jtreuting: 23 ;)

PR #227, opened 2018-01-12
Add natural ids to the reporting_olap organization views

wtritch:
Since we need to return organization natural ids for the CSV download of aggregate reports I need it in the organization views.


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -358,6 +358,7 @@ CREATE VIEW state_subject_grade_school_year(organization_id, organization_name,
     -1          AS id,
     'State'     AS name,
     'State'     AS organization_type,
+    null        AS organization_natural_id,
```
wtritch: I'm assuming a null state "natural id" makes sense.

jtreuting: makes sense to me

malaffoon: shouldn't there be a change a couple lines up too?

PR #230, opened 2018-01-17
Changes to support date range to percentile

chdietz:
Added start and end date.
Added unique index on asmt_id, start_date, end_date
- as mentioned below business logic will have to look for overlapping percentile start and end dates.
- changed min and max score in percentile_score to float
- added min and max score to percentile that represents the assessment.

Forgot to add i'm going to do some tests, maybe a PR with some tests tonight / tomorrow morning before merging

### DIFF
--- a/reporting/sql/V1_1_0_3__percentile_changes.sql
+++ b/reporting/sql/V1_1_0_3__percentile_changes.sql
@@ -8,23 +8,23 @@ USE ${schemaName};
 ALTER TABLE percentile
   ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
   ADD COLUMN end_date DATE NOT NULL AFTER start_date,
-  ADD COLUMN min_score FLOAT AFTER standard_deviation,
-  ADD COLUMN max_score FLOAT AFTER min_score,
+  ADD COLUMN min_score FLOAT NOT NULL AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT NOT NULL AFTER min_score,
   ADD UNIQUE INDEX idx__percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
 
 ALTER TABLE percentile_score
   CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
-  MODIFY COLUMN min_score FLOAT NOT NULL,
-  MODIFY COLUMN max_score FLOAT NOT NULL;
+  CHANGE COLUMN min_score min_inclusive FLOAT NOT NULL,
+  CHANGE COLUMN max_score max_exclusive FLOAT NOT NULL;
 
 ALTER TABLE staging_percentile
   ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
   ADD COLUMN end_date DATE NOT NULL AFTER start_date,
-  ADD COLUMN min_score FLOAT AFTER standard_deviation,
-  ADD COLUMN max_score FLOAT AFTER min_score,
+  ADD COLUMN min_score FLOAT NOT NULL AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT NOT NULL AFTER min_score,
   ADD UNIQUE INDEX idx__staging_percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
 
 ALTER TABLE staging_percentile_score
   CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
-  MODIFY COLUMN min_score FLOAT NOT NULL,
-  MODIFY COLUMN max_score FLOAT NOT NULL;
+  CHANGE COLUMN min_score min_inclusive FLOAT NOT NULL,
+  CHANGE COLUMN max_score max_exclusive FLOAT NOT NULL;
--- a/reporting/sql/V1_1_0_3__percentile_changes.sql
+++ b/reporting/sql/V1_1_0_3__percentile_changes.sql
@@ -1,0 +1,30 @@
+-- Add support for percentile start and end date.
+-- add min and max score to percentile for optionally supplied min and max of the assessment.
+-- change percent to percentile_rank as a more appropriate name.
+-- modify the min and max score of the rank to float.
+
+USE ${schemaName};
+
+ALTER TABLE percentile
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
+  ADD COLUMN min_score FLOAT AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT AFTER min_score,
+  ADD UNIQUE INDEX idx__percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
+
+ALTER TABLE percentile_score
+  CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
+  MODIFY COLUMN min_score FLOAT NOT NULL,
+  MODIFY COLUMN max_score FLOAT NOT NULL;
+
+ALTER TABLE staging_percentile
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
+  ADD COLUMN min_score FLOAT AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT AFTER min_score,
+  ADD UNIQUE INDEX idx__staging_percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
+
+ALTER TABLE staging_percentile_score
+  CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
+  MODIFY COLUMN min_score FLOAT NOT NULL,
+  MODIFY COLUMN max_score FLOAT NOT NULL;
--- a/reporting/sql/V1_1_0_3__percentile_changes.sql
+++ b/reporting/sql/V1_1_0_3__percentile_changes.sql
@@ -14,6 +14,7 @@ ALTER TABLE percentile
 
 ALTER TABLE percentile_score
   CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
+  ADD COLUMN score float NOT NULL AFTER percentile_rank,
   CHANGE COLUMN min_score min_inclusive FLOAT NOT NULL,
   CHANGE COLUMN max_score max_exclusive FLOAT NOT NULL;
 
@@ -26,5 +27,6 @@ ALTER TABLE staging_percentile
 
 ALTER TABLE staging_percentile_score
   CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
+  ADD COLUMN score float NOT NULL AFTER percentile_rank,
   CHANGE COLUMN min_score min_inclusive FLOAT NOT NULL,
   CHANGE COLUMN max_score max_exclusive FLOAT NOT NULL;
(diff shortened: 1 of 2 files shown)


reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,30 @@
+-- Add support for percentile start and end date.
+-- add min and max score to percentile for optionally supplied min and max of the assessment.
+-- change percent to percentile_rank as a more appropriate name.
+-- modify the min and max score of the rank to float.
+
+USE ${schemaName};
+
+ALTER TABLE percentile
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
```
chdietz: Dates so a percentile table for an assessment can be for a quarter or any date range.
Will require every table to have a date range.

reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,30 @@
...
+  ADD COLUMN min_score FLOAT AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT AFTER min_score,
```
chdietz: This is the min and max scale score for the assessment, if not provided will fall back to configuration min and max.

reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,30 @@
...
+  ADD UNIQUE INDEX idx__percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
+
+ALTER TABLE percentile_score
+  CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
+  MODIFY COLUMN min_score FLOAT NOT NULL,
+  MODIFY COLUMN max_score FLOAT NOT NULL;
```
chdietz: Changed the min and max of the percentile_rank to float.  It is possible for two ranks to have a score delta of less than one.

warehouse/sql/V1_1_0_7__percentile_dates.sql
```diff
@@ -0,0 +1,18 @@
...
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
+  ADD COLUMN min_score FLOAT AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT AFTER min_score,
+  ADD UNIQUE INDEX idx__percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
+
+ALTER TABLE percentile_score
+  CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
+  MODIFY COLUMN min_score FLOAT NOT NULL,
+  MODIFY COLUMN max_score FLOAT NOT NULL;
```
chdietz: warehouse changes are similar to reporting.  Was considering flattening the two tables in reporting but will wait to API flushed out, I suspect we do not need it.

warehouse/sql/V1_1_0_7__percentile_dates.sql
```diff
@@ -0,0 +1,18 @@
...
+-- add min and max score to percentile for optionally supplied min and max of the assessment.
+-- change percent to percentile_rank as a more appropriate name.
+-- modify the min and max score of the rank to float.
+
+USE ${schemaName};
+
+ALTER TABLE percentile
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
+  ADD COLUMN min_score FLOAT AFTER standard_deviation,
```
chdietz: notice we have `min_score` and `max_score` in `percentile` and `percentile_rank`  They mean two different things.  Anybody have an issue with that?  If yes please make a suggestion.

reporting/sql/V1_1_0_3__percentile_changes.sql
chdietz: Just realized this min and max should be not null because the optional is in import from CSV, will always be populated to imported or configuration.  - will change

reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,30 @@
...
+-- modify the min and max score of the rank to float.
+
+USE ${schemaName};
+
+ALTER TABLE percentile
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
+  ADD COLUMN min_score FLOAT AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT AFTER min_score,
+  ADD UNIQUE INDEX idx__percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
```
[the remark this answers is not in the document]
malaffoon: I wonder if we need this in reporting. The actual data integrity will be controlled in the warehouse. I guess this might help with performance.

warehouse/sql/V1_1_0_7__percentile_dates.sql
malaffoon: My only concern is that in the rank table it isn't actually min-score and max-score it is more like begin/end where the end value is exclusive. 
```
rank    min    max
  5  1500.0 2354.2
 10  2354.2 2417.8
...
```
So, for rank 5 the "max" is really an exclusive value: a score of 2354.2 is rank 10. But i guess there really isn't a good way of saying min-inclusive / max-exclusive.

reporting/sql/V1_1_0_3__percentile_changes.sql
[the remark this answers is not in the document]
chdietz: Fair point about the integrity comes from warehouse, but we have a similar unique in:
`student_group`  `idx__student_group__school_name_year`
in both reporting and warehouse. Still, just because we did it before does not make it correct ;)

As for performance, there will be a query that looks by asmt_id equal to, start and end dates by range.

warehouse/sql/V1_1_0_7__percentile_dates.sql
chdietz: hmm, thinking about it, if we have a better name I am all for it

reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,32 @@
...
+-- modify the min and max score of the rank to float.
+
+USE ${schemaName};
+
+ALTER TABLE percentile
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
+  ADD COLUMN min_score FLOAT NOT NULL AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT NOT NULL AFTER min_score,
+  ADD UNIQUE INDEX idx__percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
```
chdietz: added not null on min and max, if not provided in csv processor will fill in with config values.

reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,32 @@
...
+
+ALTER TABLE percentile_score
+  CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
+  ADD COLUMN score float NOT NULL AFTER percentile_rank,
+  CHANGE COLUMN min_score min_inclusive FLOAT NOT NULL,
```
chdietz: added score as needed for view of percentile table from UI

reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,32 @@
...
+ALTER TABLE percentile
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
+  ADD COLUMN min_score FLOAT NOT NULL AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT NOT NULL AFTER min_score,
+  ADD UNIQUE INDEX idx__percentile__asmt_start_date_end_date (asmt_id, start_date, end_date);
+
+ALTER TABLE percentile_score
+  CHANGE COLUMN percent percentile_rank TINYINT NOT NULL,
+  ADD COLUMN score float NOT NULL AFTER percentile_rank,
```
chdietz: added score as needed for view of percentile table from UI

reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,32 @@
...
+  CHANGE COLUMN min_score min_inclusive FLOAT NOT NULL,
+  CHANGE COLUMN max_score max_exclusive FLOAT NOT NULL;
```
chdietz: a better name

reporting/sql/V1_1_0_3__percentile_changes.sql
```diff
@@ -0,0 +1,32 @@
...
+-- change percent to percentile_rank as a more appropriate name.
+-- modify the min and max score of the rank to float.
+
+USE ${schemaName};
+
+ALTER TABLE percentile
+  ADD COLUMN start_date DATE NOT NULL AFTER asmt_id,
+  ADD COLUMN end_date DATE NOT NULL AFTER start_date,
+  ADD COLUMN min_score FLOAT NOT NULL AFTER standard_deviation,
+  ADD COLUMN max_score FLOAT NOT NULL AFTER min_score,
```
agorina: Are those inclusive or exclusive?

chdietz: These are the min and max score possible on the assessment.  
* same as `sbac.scaleScore.min` and `sbac.scaleScore.min`
* if not provided in CSV we fall back to this configuration.
The `min_score` is used for the first `percentile_score.min_inclusive`
The `max_score` is used for the last `percentile_score.max_exclusive`

reporting/sql/V1_1_0_3__percentile_changes.sql
chdietz: @malaffoon @agorina @jtreuting Thoughts on this?  

Just realized last night with some sample queries that this date will be compared to exam.completed_at that is a timestamp.  We really are asking for a calendar date, so it does not feel correct to store in warehouse as timestamp.

jtreuting: can't we convert the date from the CSV into a timestamp to store so that we are comparing timestamps in the query?

chdietz: Well here is a sample query see the start_date and end_date.

Two things. 
1. If we are given a calendar date, at least in the warehouse I fell it would be better to store it as a `DATE` maybe in reporting convert to timestamp if needed.
2. The other issue is what is the time portion if we convert to date.  `end_date` may need to be next day at zero hour?  an then query `<` vs `<=`

```sql
 SELECT
  e.id AS exam_id, e.asmt_id, e.oppId, e.scale_score, e.completed_at, p.id AS percentile_id, p.start_date, ps.percentile_rank
FROM exam e
LEFT JOIN percentile p ON e.asmt_id = p.asmt_id
                          AND e.completed_at >= p.start_date
                          AND e.completed_at <= p.end_date
LEFT JOIN percentile_score ps ON ps.percentile_id = p.id
                                 AND e.scale_score >= ps.min_inclusive
                                 AND e.scale_score < ps.max_exclusive;
```

malaffoon: I don't think it matters a lot. I'm fine with changing these fields to dates if that is what we expect (and specify as the format for the CSV column). The query to pick the correct row can use `>` for the start and we might have to coerce something for the end.

chdietz: @malaffoon it is `date` you mean change to `timestamp` ?

malaffoon: Uhmm, i was agreeing with this change so i meant `date`.

PR #238, opened 2018-01-26
Add a user_report_metadata table for holding semi-structured metadata…

wtritch:
… about a report

We want an AggregateReport type-specific report property (totalReportRows) in addition two existing PDF report type-specific report properties (totalChunkCount, completedChunkCount).

As the third type-specific report property, I'm breaking out a metadata table that allows for arbitrary name/value metadata properties to be added to a report.


malaffoon:
I'm fine with this change. Just curious though, aren't we already storing json blobs with report "metadata"? Are these values that need to be queried/set independently?

wtritch:
@malaffoon we're storing the immutable "request" as a json blob.  This is more for storing processing/response properties.

PR #239, opened 2018-01-29
Migrate user_report chunk count data to metadata.

wtritch:


reporting/sql/V1_1_0_4__user_report_metadata.sql
```diff
@@ -1,5 +1,7 @@
 -- Add user report metadata to contain report-type-specific information
 
+USE ${schemaName};
```
wtritch: shoot.  I missed this on the last PR.  This may cause us migration issues.

PR #240, opened 2018-01-30
add percentile content type

chdietz:
add NORMS code to `import_content` table.

there will be related change in RDW_Common enum

### DIFF
--- a/warehouse/sql/V1_1_0_8__percentile_content_type.sql
+++ b/warehouse/sql/V1_1_0_8__percentile_content_type.sql
@@ -1,0 +1,5 @@
+-- Add percentile content type
+
+USE ${schemaName};
+
+INSERT INTO import_content (id, name) VALUES (7, 'PERCENTILE');
--- a/warehouse/sql/V1_1_0_8__percentile_content_type.sql
+++ b/warehouse/sql/V1_1_0_8__percentile_content_type.sql
@@ -1,5 +1,5 @@
--- Add percentile content type
+-- Add NORMS content type
 
 USE ${schemaName};
 
-INSERT INTO import_content (id, name) VALUES (7, 'PERCENTILE');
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_1_0_8__percentile_content_type.sql
```diff
@@ -0,0 +1,5 @@
+-- Add percentile content type
+
+USE ${schemaName};
+
+INSERT INTO import_content (id, name) VALUES (7, 'PERCENTILE');
```
chdietz: matches enum in RDW_Common

malaffoon: Guess it doesn't matter a lot, but can we make it `NORMS` or `NORMATIVE_DATA` instead of `PERCENTILE`? Percentiles are just one part of it.

chdietz: We could, but the percentiles are for the normative group of data, we are not loading the normative data set we are loading the ranks for the normative data, correct?

malaffoon: This is the import content ... it includes percentiles, average scale score, std.err., etc.

chdietz: If we agree that `Normative` is a better term, I think `Norm` is an abbreviation for `Normative`.  And we do not append `Data` to other terms is the set of content types
so I would suggest `NORMATIVE`

warehouse/sql/V1_1_0_8__percentile_content_type.sql
```diff
@@ -0,0 +1,5 @@
+-- Add NORMS content type
+
+USE ${schemaName};
+
+INSERT INTO import_content (id, name) VALUES (7, 'NORMS');
```
chdietz: changed to NORMS

PR #241, opened 2018-02-02
Add coded to asmt table.

agorina:


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -245,6 +245,7 @@ CREATE TABLE state_embargo (
 CREATE TABLE asmt (
   id int encode raw NOT NULL PRIMARY KEY,
   grade_id smallint NOT NULL,
+  grade_code character varying(2) NOT NULL,
```
agorina: People really like codes...

malaffoon: but only for grade, not subject, type, etc?

agorina: Good point. If I add `subject_code`, I can remove one more JOIN. Will do.

agorina:
This was not such a good idea.

Just to explain why: the report SQL uses ids in WHERE and GROUP BY. By adding codes, we now have to add them to the GROUP BY, that leads to changes to the SORT KEY.

After talking with @wtritch , I am going to abandon this change and will do translation in Java.

PR #246, opened 2018-03-04
POC queries for Longitudinal Aggregate report - round 2

agorina:
As I said before - this is to solicit feedback. I am not planning on merging it in here.


reporting_olap/sql/_OlapEntitiesCohortSetup.sql
```diff
@@ -0,0 +1,87 @@
+INSERT INTO grade (id, code) VALUES (-3,'-3'),(-4 ,'-4'),(-5, '-5'),(-6, '-6');
```
agorina: Sample data.

reporting_olap/sql/_longitudinal_sample.sql
```diff
@@ -0,0 +1,134 @@
+-- -------------------------------------------------------------------------------------------------
+-- School level report
+-- -------------------------------------------------------------------------------------------------
+SELECT
+  SUM(CASE WHEN years_in_one_org = 3 THEN 1 ELSE 0 END) AS cohort_size,
+  count(*) AS total_size,
+  avg(CASE WHEN years_in_one_org = 3 THEN scale_score ELSE NULL END) AS cohort_avg,
```
agorina: As was mentioned before:
All aggregate functions but `count` ignore `NULL`.  This allows to return both cohort and general population values in one statement.  We will need to do some more calculations in Java (for the aggregate std err).

reporting_olap/sql/_longitudinal_sample.sql
```diff
@@ -0,0 +1,134 @@
...
+           OR (fe.school_year = 1999 AND a.grade_id = -5)
+           -- pad to X years?
+           OR (fe.school_year = -1 AND a.grade_id = -1)
+           OR (fe.school_year = -2 AND a.grade_id = -2)
+         )
+         AND school_id IN (-9)
+       -- AND other filters here, including embargo one
+     ) cfe
+-- Alternatively other filters could be here. This would make the query builder simpler since it would resemble the current query more,
+--  but it could be slower.  Worth researching? I think this will allow for re-using the existing query builder code.
```
agorina: My goal is to be able to reuse the current query building logic we have for the custom aggregate report.

reporting_olap/sql/_longitudinal_sample.sql
```diff
@@ -0,0 +1,134 @@
...
+FROM (
+       -- school cohort
+       SELECT
+         fe.scale_score,
+         fe.asmt_id,
+         fe.school_year,
+         fe.school_id,
+         -- add dimension column if not Overall, alternatively list all; will listing all make it slower?
+         count(*) OVER (PARTITION BY student_id, school_id ) AS years_in_one_org
+       FROM fact_student_exam fe
```
agorina: Can we reuse the same fact table? It has all the data, but would about indexes?

My preliminary research shows that for this report the simplest query plan requires a sort key to be `COMPOUND SORTKEY (student_id, school_id)` due to `PARTITION BY student_id, school_id`.

We have a JIRA to revisit the current SORTKEY key on this table. 
This requires a larger discussion that this PR, so I will stop at this and we can all chat later.

reporting_olap/sql/_longitudinal_sample.sql
```diff
@@ -0,0 +1,134 @@
...
+         a.subject_id = 1
+         AND a.type_id = 1
+         AND
+         (
+           (fe.school_year = 1997 AND a.grade_id = -3)
+           OR (fe.school_year = 1998 AND a.grade_id = -4)
+           OR (fe.school_year = 1999 AND a.grade_id = -5)
+           -- pad to X years?
+           OR (fe.school_year = -1 AND a.grade_id = -1)
+           OR (fe.school_year = -2 AND a.grade_id = -2)
```
agorina: This is similar to padding IN clause to X values. I have tested and it seems to have the same effect (reused cached query).

reporting_olap/sql/_longitudinal_sample.sql
```diff
@@ -0,0 +1,134 @@
...
+-- School level report
+-- -------------------------------------------------------------------------------------------------
+SELECT
+  SUM(CASE WHEN years_in_one_org = 3 THEN 1 ELSE 0 END) AS cohort_size,
+  count(*) AS total_size,
+  avg(CASE WHEN years_in_one_org = 3 THEN scale_score ELSE NULL END) AS cohort_avg,
+  avg(scale_score) AS avg,
+  stddev_samp(CASE WHEN years_in_one_org = 3 THEN scale_score ELSE NULL END) AS cohort_stdev,
+  stddev_samp(scale_score) AS stdev,
+  asmt_id,
```
agorina: Including `asmt_id` (similar to the current report). If there are two assessments for the same year, the data will look strange, but this is the goal since it should not happen.

PR #248, opened 2018-03-09
DWR-1582 Concurrency issue with exam ingest

agorina:


warehouse/sql/V1_2_0_0__student_upsert.sql
```diff
@@ -0,0 +1,72 @@
...
+          birthday                      = p_birthday,
+          inferred_school_id            = p_exam_school_id,
+          update_import_id              = p_import_id
+        WHERE id = p_id;
+      END IF;
+    ELSE
+      INSERT INTO student (ssid, last_or_surname, first_name, middle_name, gender_id, first_entry_into_us_school_at, lep_entry_at, lep_exit_at, birthday, inferred_school_id, import_id, update_import_id)
+      VALUES (p_ssid, p_last_or_surname, p_first_name, p_middle_name, p_gender_id, p_first_entry_into_us_school_at, p_lep_entry_at, p_lep_exit_at, p_birthday, p_exam_school_id, p_import_id, p_import_id);
+
+      SELECT id, 2 INTO p_id, p_updated FROM student WHERE ssid = p_ssid;
```
agorina: This is the line that has changed. 

In the first release it was similar to this one and on some occasions it would return NULL, and since we have a retry on the repository it would retry and works. (I still do not know why it would return null, but it did).

We have changed it to `SELECT LAST_INSERT_ID(), 2 INTO p_id, p_updated;` And now instead of NULL it was returning the last inserted id of who knows what.

PR #251, opened 2018-03-24
ELAS

malaffoon:
I won't be merging this as-is but i wanted to address some questions i had and this seemed the easiest way ...

### DIFF
--- a/reporting/sql/v1_2_0_0__elas_gender.sql
+++ b/reporting/sql/v1_2_0_0__elas_gender.sql
@@ -1,0 +1,37 @@
+-- Flyway script to add ELAS (English Language Acquisition Status) and unknown gender
+--
+-- NOTE: instead of migrating CODES, which would require multiple scripts, just make the changes directly.
+
+USE ${schemaName};
+
+INSERT INTO gender (id, code) VALUES
+  (3, 'NonBinary');
+
+CREATE TABLE IF NOT EXISTS elas (
+  id tinyint NOT NULL PRIMARY KEY,
+  code varchar(20) NOT NULL UNIQUE
+);
+
+INSERT INTO elas (id, code) VALUES
+  (1, 'EO'),
+  (2, 'EL'),
+  (3, 'IFEP'),
+  (4, 'RFEP'),
+  (5, 'TBD');
+
+CREATE TABLE IF NOT EXISTS staging_elas (
+  id tinyint NOT NULL PRIMARY KEY,
+  code varchar(20) NOT NULL UNIQUE
+);
+
+
+ALTER TABLE exam
+  MODIFY COLUMN lep TINYINT NULL,
+  ADD COLUMN elas_id TINYINT NULL,
+  ADD COLUMN elas_code VARCHAR(20) NULL,
+-- ??? i don't see FKs for completeness_id or administration_condition_id, do we want these?
+  ADD CONSTRAINT fk__exam__elas FOREIGN KEY (elas_id) REFERENCES elas(id);
+
+ALTER TABLE staging_exam
+  MODIFY COLUMN lep TINYINT NULL,
+  ADD COLUMN elas_id TINYINT NULL;
--- a/reporting/sql/v1_2_0_0__elas_gender.sql
+++ b/reporting/sql/v1_2_0_0__elas_gender.sql
@@ -1,11 +1,17 @@
 -- Flyway script to add ELAS (English Language Acquisition Status) and unknown gender
 --
+-- ELAS: this is a new attribute for students. It overlaps with LEP but will be treated as
+-- an independent, optional attribute. LEP will be changed to be optional as well.
+--
+-- As part of this change, remove the ids for denormalized exam/student fields where only the code is used.
+-- This includes elas, completeness and administration_condition.
+--
 -- NOTE: instead of migrating CODES, which would require multiple scripts, just make the changes directly.
 
 USE ${schemaName};
 
 INSERT INTO gender (id, code) VALUES
-  (3, 'NonBinary');
+  (3, 'Nonbinary');
 
 CREATE TABLE IF NOT EXISTS elas (
   id tinyint NOT NULL PRIMARY KEY,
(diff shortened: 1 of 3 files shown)


reporting/sql/v1_2_0_0__elas_gender.sql
```diff
@@ -0,0 +1,37 @@
...
+  id tinyint NOT NULL PRIMARY KEY,
+  code varchar(20) NOT NULL UNIQUE
+);
+
+
+ALTER TABLE exam
+  MODIFY COLUMN lep TINYINT NULL,
+  ADD COLUMN elas_id TINYINT NULL,
+  ADD COLUMN elas_code VARCHAR(20) NULL,
+-- ??? i don't see FKs for completeness_id or administration_condition_id, do we want these?
```
malaffoon: @agorina ?

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -299,6 +310,8 @@ CREATE TABLE fact_student_exam (
   school_year smallint encode raw NOT NULL,
   iep smallint encode lzo NOT NULL,
   lep smallint encode lzo NOT NULL,
+  -- ??? it seems we don't want NULL fields so what should be done here? add an 'unknown' value? or is null okay?
```
malaffoon: @agorina ?

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -316,6 +329,9 @@ CREATE TABLE fact_student_exam (
   CONSTRAINT fk__fact_student_exam__student FOREIGN KEY(student_id) REFERENCES student(id),
   CONSTRAINT fk__fact_student_exam__iep FOREIGN KEY(iep) REFERENCES strict_boolean(id),
   CONSTRAINT fk__fact_student_exam__lep FOREIGN KEY(lep) REFERENCES strict_boolean(id),
+  CONSTRAINT fk__fact_student_exam__elas FOREIGN KEY(elas_id) REFERENCES elas(id),
+  -- ??? why don't administration_condition_id, completeness_id have FKs?
+  -- ??? i don't see section504 ever being set to anything but 0/1 so why boolean and not strict_boolean?
```
malaffoon: @agorina ?

warehouse/sql/V1_2_0_0__elas_gender.sql
```diff
@@ -0,0 +1,30 @@
...
+  MODIFY COLUMN lep TINYINT NULL,
+  ADD COLUMN elas_id TINYINT NULL,
+  -- ??? i don't see FKs for completeness_id or administration_condition_id, do we want these?
+  ADD CONSTRAINT fk__exam__elas FOREIGN KEY (elas_id) REFERENCES elas(id);
+
+#
+# -- trigger a CODES migration
+# -- NOTE: this just catches the
+# INSERT INTO import(status, content, contentType, digest) VALUES
+#   (1, 3, 'v1.2 update', 'v1.2 update');
```
malaffoon: I will delete this. To use migrate to catch these changes i'd have to have two scripts deployed separately: the first one creates the elas table and populates it so that that migrate code changes can be deployed, then the second one would just create this import record to trigger the migration. Seems silly.

reporting/sql/v1_2_0_0__elas_gender.sql
agorina: The comment in V1_0_0_0__ddl.sql says: _"intentionally not constrained since it is denormalized"_, but we can revisit it. 

I think we may want to remove the ids since the application is not using them.

reporting_olap/sql/V1_0_0_0__ddl.sql
agorina: You are right, we do not want NULL here and we need to add 'unknown'.

reporting_olap/sql/V1_0_0_0__ddl.sql
agorina: _"why don't administration_condition_id, completeness_id have FKs"_ - hmmm...I think the 
reason is historical.

From Redshift: _"Define primary key and foreign key constraints between tables wherever appropriate. Even though they are informational only, the query optimizer uses those constraints to generate more efficient query plans.
Do not define primary key and foreign key constraints unless your application enforces the constraints. Amazon Redshift does not enforce unique, primary-key, and foreign-key constraints."_

Since we never JOIN on these ids I did not think the FKs are needed. But we now changed the SQL to not JOIN on most of them. So I guess we could/should either remove most of them or add all of them.

One thing I noticed is that Redshift seems to enforce a FK on DELETE. This maybe a benefit of adding all of them? (if you decide to add it, please test that I am not lying here about DELETE).

About '504' - I think there was a requirement to make it NULLable. I will try to remember where it came from and will post my finding here

warehouse/sql/V1_2_0_0__elas_gender.sql
```diff
@@ -0,0 +1,30 @@
...
+  (1, 'EO'),
+  (2, 'EL'),
+  (3, 'IFEP'),
+  (4, 'RFEP'),
+  (5, 'TBD');
+
+ALTER TABLE exam
+  MODIFY COLUMN lep TINYINT NULL,
+  ADD COLUMN elas_id TINYINT NULL,
+  -- ??? i don't see FKs for completeness_id or administration_condition_id, do we want these?
```
agorina: For the performance reasons we took an approach of not enforcing most of the FKs in `warehouse` since Jave enforces it. But it was at the very beginning of the project. If you think we should revisit it, let's chat.

reporting/sql/v1_2_0_0__elas_gender.sql
```diff
@@ -0,0 +1,41 @@
...
+CREATE TABLE IF NOT EXISTS staging_elas (
+  id tinyint NOT NULL PRIMARY KEY,
+  code varchar(20) NOT NULL UNIQUE
+);
+
+ALTER TABLE exam
+  MODIFY COLUMN lep TINYINT NULL,
+  ADD COLUMN elas_code VARCHAR(20) NULL,
+  DROP COLUMN completeness_id,
+  DROP COLUMN administration_condition_id;
```
malaffoon: (sort of) unrelated: these id's aren't used in reporting so remove 'em

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -299,13 +310,14 @@ CREATE TABLE fact_student_exam (
...
+  elas_id smallint encode lzo NOT NULL,
   section504 smallint encode lzo NOT NULL,
   economic_disadvantage smallint encode lzo NOT NULL,
   migrant_status smallint encode lzo NOT NULL,
   completeness_id smallint encode lzo NOT NULL,
   administration_condition_id smallint encode lzo NOT NULL,
-  scale_score float NOT NULL encode bytedict ,
-  performance_level smallint NOT NULL encode lzo,
+  scale_score float encode bytedict NOT NULL,
+  performance_level smallint encode lzo NOT NULL,
```
malaffoon: unrelated, just making order consistent

malaffoon:
I don't have an approval but i need this change to move forward; i'll address any further comments (if i can).

reporting/sql/v1_2_0_0__elas_gender.sql
malaffoon: Correct, i checked. It is used in OLAP but not individual reporting.

PR #255, opened 2018-04-09
add elas start date to reporting

malaffoon:


reporting/sql/V1_2_0_2__elas_date.sql
```diff
@@ -0,0 +1,10 @@
+-- Flyway script to add ELAS (English Language Acquisition Status) date
+--
+
+USE ${schemaName};
+
+ALTER TABLE exam
+  ADD COLUMN elas_start_at DATE NULL;
```
malaffoon: There is no way to populate this from the existing data.

PR #256, opened 2018-04-10
DWR-1652 Modify fact table for longitudinal report

agorina:


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -119,6 +119,8 @@ CREATE TABLE staging_exam (
   id bigint NOT NULL PRIMARY KEY,
   student_id int NOT NULL,
   grade_id smallint NOT NULL,
+  subject_id smallint,    -- this is updated later, not as part of loading into the staging
+  asmt_grade_id smallint, -- this is updated later, not as part of loading into the staging
```
agorina: When updating `fact_student_exam_longitudinal` I kept running into :
`java.sql.SQLException: [Amazon](500310) Invalid operation: Target table must be part of an equijoin predicate;`. I have tried to rewrite it as suggested by AWS, but could not get it to work. 

To work around it, I am adding these two fields to the staging table and updating them in the migrate process. This avoids a JOIN in the UPDATE.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -395,19 +399,22 @@ CREATE TABLE fact_student_exam_longitudinal (
...
-  CONSTRAINT fk__fact_student_exam__iep FOREIGN KEY(iep) REFERENCES strict_boolean(id),
-  CONSTRAINT fk__fact_student_exam__lep FOREIGN KEY(lep) REFERENCES strict_boolean(id),
-  CONSTRAINT fk__fact_student_exam__elas FOREIGN KEY(elas_id) REFERENCES elas(id),
-  CONSTRAINT fk__fact_student_exam__completeness FOREIGN KEY(completeness_id) REFERENCES completeness(id),
-  CONSTRAINT fk__fact_student_exam__administration_comdition FOREIGN KEY(administration_condition_id) REFERENCES administration_condition(id),
-  CONSTRAINT fk__fact_student_exam__section504 FOREIGN KEY(section504) REFERENCES boolean(id),
-  CONSTRAINT fk__fact_student_exam__economic_disadvantage FOREIGN KEY(economic_disadvantage) REFERENCES strict_boolean(id),
-  CONSTRAINT fk__fact_student_exam__migrant_status FOREIGN KEY(migrant_status) REFERENCES boolean(id)
-)  COMPOUND SORTKEY (student_id, school_year, asmt_id, school_id);
+  CONSTRAINT fk__fact_student_exam_longitudinal__asmt FOREIGN KEY(asmt_id) REFERENCES asmt(id),
```
agorina: Fixed a typo.

reporting_olap/sql/V1_0_0_0__ddl.sql
malaffoon: Not grokking all of this. Why can't you set those values as part of loading into staging?

agorina: We can do it if we do the JOIN during the extract from Aurora. My philosophy was that both extract from Aurora and load to the final destination should be as simple as they could be. While all data manipulation should happen in the staging. 

But I agree that this is not a big change to add a JOIN from `exam` to `asmt` on the extract. If it feels cleaner, I can make this change.

PR #257, opened 2018-04-12
Add exam index to support IAB dashboard query.

agorina:


reporting/sql/V1_2_0_3__iab_dashboard_exam_index.sql
```diff
@@ -0,0 +1,13 @@
+-- Flyway script to add exam index for the IAB dashboard query
+--
+USE ${schemaName};
+
+ALTER TABLE exam
+  -- new index for the dashboard query
+  ADD INDEX idx__exam__student_asmt_type_school_year_completed_at (student_id, asmt_id, type_id, school_year, completed_at),
```
agorina: Here is how the SQL looks:
``` 
SELECT
  count(*),
  a.label as asmt_label,
  g.code as grade_code,
  asmt_id as asmt_id,
  max(completed_at),
  round(avg(scale_score)) AS score,
  round(stddev_samp(scale_score)/sqrt(count(*))) as std_err,
  sum(CASE WHEN performance_level = 1 THEN 1 ELSE 0 END) AS level1,
  sum(CASE WHEN performance_level = 2 THEN 1 ELSE 0 END) AS level2,
  sum(CASE WHEN performance_level = 3 THEN 1 ELSE 0 END) AS level3
FROM (
  -- latest IAB exam and completed_at per student, iab and school year
  SELECT
    e1.asmt_id,
    e1.completed_at,
    e1.scale_score,
    e1.scale_score_std_err,
    e1.performance_level,
    e1.school_id,
    e1.student_id
  FROM exam AS e1
    LEFT OUTER JOIN exam AS e2 ON e1.student_id = e2.student_id
                                  AND e1.asmt_id = e2.asmt_id
                                  AND e1.school_year = e2.school_year
                                  AND e1.type_id = 2 -- IABs only
                                  AND (e1.completed_at < e2.completed_at OR (e1.completed_at = e2.completed_at AND e1.id < e2.id))
  WHERE e2.student_id IS NULL
        AND e1.school_year = :school_year
        AND e1.student_id IN(
            SELECT student_id
              FROM student_group_membership gm JOIN  user_student_group usg ON usg.student_group_id = gm.student_group_id
              WHERE gm.student_group_id = :group_id AND usg.user_login = :user_login
        )
) e
  JOIN asmt a ON a.id = e.asmt_id
  JOIN grade g ON g.id = a.grade_id
  JOIN school s ON e.school_id=s.id
  JOIN student st ON st.id=e.student_id
  JOIN school isch ON st.inferred_school_id = isch.id
WHERE
  ${sql.snippet.studentExamPermissionWithoutEmbargo} -- this is for IAB only, embargo is not applicable to IABs
GROUP BY asmt_id;
```

reporting/sql/V1_2_0_3__iab_dashboard_exam_index.sql
```diff
@@ -0,0 +1,13 @@
...
+
+  -- replace foreign key index with the new one
+  DROP FOREIGN KEY fk__exam__student,
+  DROP INDEX idx__exam__student;
+
+ALTER TABLE exam ADD CONSTRAINT fk__exam__student FOREIGN KEY (student_id) REFERENCES student(id);
```
agorina: This is now redundant since the FK can now use the new index.

malaffoon: i'm confused, what was it before?

agorina: The FK was exactly the same, the change is what index it is using. Before it was using `idx__exam__student`. But since now I have introduced another index that starts with the 'student_id' it will use it. From MySQL:

_"InnoDB requires indexes on foreign keys and referenced keys so that foreign key checks can be fast and not require a table scan. In the referencing table, there must be an index where the foreign key columns are listed as the first columns in the same order. Such an index is created on the referencing table automatically if it does not exist."_

agorina: It has been taking a while (still do not know how long) to recreate this FK in AWS Dev. I am debating this change. 

I will post the time once it is done here and we can discuss if this change is worth doing.

reporting/sql/V1_2_0_3__iab_dashboard_exam_index.sql
jtreuting: does it seem like it performs okay (not under load yet of course)?

agorina: Yes, so far it looks good, almost instantaneous. 

As expected, the outer step is done "Using temporary; Using filesort". So as you mentioned, it needs to be tested under load.

agorina: I have created a JIRA to test this under load.

PR #263, opened 2018-04-25
Add/modify warehouse schema to allow ingesting configurable subject xml

wtritch:
This PR contains the new tables and table modifications to support ingesting subject XML into warehouse.

I'm not sure exactly what's required for auditing.  I likely need to extend the warehouse.subject table with created/updated/import_id/update_import_id columns?


warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+-- Table for holding subject configurations in the context of an assessment type
+CREATE TABLE IF NOT EXISTS asmt_type_subject (
+  asmt_type_id TINYINT,
+  subject_id TINYINT,
+  performance_levels TINYINT,
+  performance_level_standard_cutoff TINYINT NULL,
+  sub_score_performance_levels TINYINT,
+  sub_score_performance_level_standard_cutoff TINYINT NULL,
+  created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
wtritch: I'm not sure "updated" should be here, although I suppose it could be... will we need audit_xxx tables for this since values are technically modifiable when ingesting an updated subject xml?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+  id SMALLINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  subject_id TINYINT,
+  parent_id SMALLINT NULL,
+  code VARCHAR(10),
+  code_hierarchy VARCHAR(100), -- Looks like "1-1|1-LT|SOCK_R" for an ELA target, should we also store the reverse for easy hierarchy loading?
+  created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  CONSTRAINT fk__item_category__parent FOREIGN KEY (parent_id) REFERENCES item_category(id) ON DELETE CASCADE,
+  CONSTRAINT fk__item_category__subject FOREIGN KEY (subject_id) REFERENCES subject(id),
+  UNIQUE INDEX idx__item_category__code_hierarchy(code_hierarchy)
+);
```
wtritch: This table replaces the claim and target tables with parent_id self-references.  For faster lookups, I'm thinking we can use the code_hierarchy column so that when we get a target id like "1|AGP-A|A" we can find the "target" category quickly.  Thoughts?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+  scale_score SMALLINT,
+  scale_score_std_err FLOAT,
+  performance_level TINYINT,
+  theta_score FLOAT,
+  theta_score_std_err FLOAT,
+  created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY(exam_id, category_id),
+  CONSTRAINT fk__exam_sub_scores__exam FOREIGN KEY (exam_id) REFERENCES exam(id),
+  CONSTRAINT fk__exam_sub_scores__item_category FOREIGN KEY (category_id) REFERENCES item_category(id)
+);
```
wtritch: By referencing a "category_id" this table can hold both "claim" and "target" scores.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+
+-- Add category references to assessment items
+ALTER TABLE item
+  ADD COLUMN category_id SMALLINT NULL, -- This should reference the target (leaf) category
+  ADD CONSTRAINT fk__item__item_category FOREIGN KEY (category_id) REFERENCES item_category(id);
```
wtritch: Rather than binding an item to both a "claim" and a "primary target" I'm instead going to bind it to a "primary category" and we can get the claim from the category hierarchy... thoughts?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
+-- Prepare for configurable subjects by creating the tables required to ingest and store
+-- subject xml payloads.
+-- Question: auditing?  What's required?
```
malaffoon: I don't see why subject would be audited. Auditing is focused on PII.

wtritch: You just made my day.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+
+USE ${schemaName};
+
+-- Warehouse translation table for holding English message translations provided
+-- with the subject XML
+CREATE TABLE IF NOT EXISTS translation (
```
malaffoon: i know you're following convention but i really don't like the `IF NOT EXISTS` part of these. We should know if the table is supposed to exist when this script is run. If it exists, we have a problem and the script shouldn't continue. If it doesn't exist, well we don't need that part of the command.

wtritch: removed

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+CREATE TABLE translation (
+  namespace VARCHAR(10),
+  label_code VARCHAR(128),
+  language_code VARCHAR(3),  -- Should this be assumed as "eng" ?
+  label TEXT,
+  PRIMARY KEY(namespace, label_code, language_code)
+);
+
+-- Table for holding subject configurations in the context of an assessment type
+CREATE TABLE asmt_type_subject (
```
malaffoon: nit: why not `subject_asmt_type`? This will be a child table off of `subject` right?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+  PRIMARY KEY(asmt_type_id, subject_id),
+  CONSTRAINT fk__asmt_type_subject__asmt_type FOREIGN KEY (asmt_type_id) REFERENCES asmt_type(id),
+  CONSTRAINT fk__asmt_type_subject__subject FOREIGN KEY (subject_id) REFERENCES subject(id)
+);
+
+-- Table for holding hierarchical item categories (replaces claim and target tables)
+CREATE TABLE item_category (
+  id SMALLINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  subject_id TINYINT,
+  parent_id SMALLINT NULL,
```
malaffoon: ugh, are we going to end up doing self-joins?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
malaffoon: i like the idea of storing the "full" code

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
wtritch: ...possibly.  But that's part of the reason I added the code_hierarchy column to allow for searching from the bottom-up.  I'm debating having an additional inverted column to allow for searching from the top-down while still using an index... but ideally the table will never be large enough that a left-wildcard search is prohibitively expensive.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
wtritch: good call, changed

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+  code VARCHAR(10),
+  code_hierarchy VARCHAR(100), -- Looks like "1-1|1-LT|SOCK_R" for an ELA target, should we also store the reverse for easy hierarchy loading?
+  created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  CONSTRAINT fk__item_category__parent FOREIGN KEY (parent_id) REFERENCES item_category(id) ON DELETE CASCADE,
```
malaffoon: Did you test this cascade delete? I've just never done a self-referential cascade.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
wtritch: ...I think this would the "full" code? Although that statement could be a little misleading with our "scorable-multi-claim" root elements.  The assessment package will reference a target like "2|AGP-A|A" but it'll be represented in this column as "AGP-A|A|2|Sock_2"  ...although we should probably pick a different separator than `|` since SBAC already uses it.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
wtritch: Nope, but you made me go googling:
```
If ON UPDATE CASCADE or ON UPDATE SET NULL recurses to update the same table it has previously updated during the cascade, it acts like RESTRICT. This means that you cannot use self-referential ON UPDATE CASCADE or ON UPDATE SET NULL operations. This is to prevent infinite loops resulting from cascaded updates. A self-referential ON DELETE SET NULL, on the other hand, is possible, as is a self-referential ON DELETE CASCADE. Cascading operations may not be nested more than 15 levels deep.
```

wtritch: So as long as we don't have 15 levels of categories, we should be fine.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+  namespace VARCHAR(10),
+  label_code VARCHAR(128),
+  language_code VARCHAR(3),  -- Should this be assumed as "eng" ?
+  label TEXT,
+  PRIMARY KEY(namespace, label_code, language_code)
+);
+
+-- Table for holding subject configurations in the context of an assessment type
+CREATE TABLE asmt_type_subject (
+  asmt_type_id TINYINT,
```
agorina: Could these be NULL? (and the next one too)

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+  subject_id TINYINT,
+  performance_levels TINYINT,
+  performance_level_standard_cutoff TINYINT NULL,
```
agorina: nit: I do not think you need 'NULL' here.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
agorina: Interesting...are we going to allow for updates of any of the fields? Or just some.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
agorina: Just curious: do we have a JIRA to create SQL script to covert production data into this new format?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
jtreuting: i would think we would save the file that was used so we have an audit trail of sorts with the files that were used being stored somewhere.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
jtreuting: I don't think this will work as the scores for targets are most likely going to be different than for claims.  Having a scale score isn't enough for target level reports (see: https://confluence.fairwaytech.com/display/SBACDW/TRT+target+scores+definitions)

So i think they made need to be stored separately.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
jtreuting: i'm trying to think if there is a hole in that logic but i think it will work.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,77 @@
...
+  language_code VARCHAR(3),  -- Should this be assumed as "eng" ?
+  label TEXT,
+  PRIMARY KEY(namespace, label_code, language_code)
+);
+
+-- Table for holding subject configurations in the context of an assessment type
+CREATE TABLE asmt_type_subject (
+  asmt_type_id TINYINT,
+  subject_id TINYINT,
+  performance_levels TINYINT,
```
jtreuting: i don't feel strongly about it either way, but at first it wasn't clear that this was just the number of performance levels and not the actual performance levels (yes i know once you see TINYINT it clears things up).  but maybe naming it `number_of_performance_levels` is more clear?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
malaffoon: Sorry i wasn't clear. Yes, the would be the full code and i was just agreeing with the approach you are taking.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
wtritch: If we ingest the same subject multiple times, any of these could potentially change.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
wtritch: Added columns to represent the target-level residual scores

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
wtritch: maybe `performance_level_count` ?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
wtritch: @agorina I'm considering that as part of the development process.  It's going to be a bit of a dance w.r.t. production.  We'll need to migrate the schema, ingest the subjects, migrate the data, then optionally clean up the schema (removing unused tables/columns)

agorina: And I am sure this is a given, but we would need to have SQL scripts that validate any data migrate. Just do not want this effort to be missed.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
agorina: I think I said it before, but the idea of updating the subject definition scares me. I think we may want to limit what is `updatable` (similar to how we do it with the assessment packages now). Not every field could be updated without a significant ripple effect.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,81 @@
...
+-- Question: auditing?  What's required?
+
+USE ${schemaName};
+
+-- Warehouse translation table for holding English message translations provided
+-- with the subject XML
+CREATE TABLE translation (
+  namespace VARCHAR(10) NOT NULL,
+  label_code VARCHAR(128) NOT NULL,
+  language_code VARCHAR(3) NOT NULL,  -- Should this be assumed as "eng" ?
```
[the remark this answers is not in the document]
wtritch: I'm going to drop the language_code column and rename the table to "label"

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,87 @@
...
+ALTER TABLE item_other_target
+  ADD COLUMN category_id SMALLINT,
+  ADD CONSTRAINT fk__item_other_target__item_category FOREIGN KEY (category_id) REFERENCES item_category(id);
+
+-- Modify subject table to act as import/migration root
+ALTER TABLE subject
+  ADD COLUMN created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  ADD COLUMN updated TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  ADD COLUMN import_id BIGINT,
+  ADD COLUMN update_import_id BIGINT;
```
wtritch: I added import columns to the subject table, since it will now be an import root element.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,87 @@
...
+USE ${schemaName};
+
+-- Warehouse label table for holding English label key/value pairs.
+-- Functionally, these should be migrated to reporting as "eng" translation values.
+CREATE TABLE label (
+  namespace VARCHAR(10) NOT NULL,
+  label_code VARCHAR(128) NOT NULL,
+  label TEXT,
+  PRIMARY KEY(namespace, label_code)
+);
```
wtritch: Hm... how should we trigger migration of these values?  Right now they will only be added via Subject ingest... but in the future?  Should each label have an associated import_id?  Should labels be a root migration element?  @agorina @malaffoon ?

agorina: Since this table does not have a FK to 'subject' it does not feel right to consider 'subject' as a root for it. 

Could anything else besides the Subject ingest update the data in this table? (outside of the manual update)?

Should this be part of the 'code' migrate?

wtritch: Theoretically, any ingest where we want to accept English text during ingest and then migrate it to reporting as a translation key/value can use this table.  Right now, only Subject will be using it, but the two aren't tightly-coupled.

Interesting point on using 'code' to migrate it... When do code migrations run?  Is it a one-time at-system-setup migration?  Because *when* we add additional subject-scoped text to the application, we'll need to update the Subject definitions and re-import them for cleanliness and to ensure standing up a new installation is smooth.  Regardless, we'll still need an `updated` column in order to avoid migrating the entire table on every triggering event, right?

malaffoon: Would migrating the entire table be a big deal?

wtritch: There will be... roughly ~2500 rows in the table with just Math and ELA labels.  While it wouldn't break, it feels a little messy.  Especially since we'll need to do some processing on the reporting end to avoid overwriting any translation values that have been manually updated more recently than they have been updated via a Subject ingest/update.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,87 @@
...
+  code VARCHAR(10) NOT NULL,
+  code_hierarchy VARCHAR(100) NOT NULL, -- Looks like "1-1|1-LT|SOCK_R" for an ELA target, should we also store the reverse for easy hierarchy loading?
+  created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  CONSTRAINT fk__item_category__parent FOREIGN KEY (parent_id) REFERENCES item_category(id) ON DELETE CASCADE,
+  CONSTRAINT fk__item_category__subject FOREIGN KEY (subject_id) REFERENCES subject(id),
+  UNIQUE INDEX idx__item_category__code_hierarchy(code_hierarchy)
+);
+
+-- Create sub-score table for holding category (claim and target) scores.
+CREATE TABLE exam_sub_scores (
```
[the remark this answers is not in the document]
agorina: I am fine with it.

If we want to be 'truly' data driven, then (as we discussed yesterday) the data should not be flattened and instead we should be able to take any 'measureOf' and 'measureLabel' values from this:

```
 <Score measureOf="Overall" measureLabel="ThetaScore" value="-0.0933834960601748" standardError="0.256414973091891"/>
  <Score measureOf="Overall" measureLabel="PerformanceLevel" value="3" standardError=""/>
```

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,92 @@
...
+ALTER TABLE claim ADD CONSTRAINT fk__claim__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+ALTER TABLE common_core_standard ADD CONSTRAINT fk__common_core_standard__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+ALTER TABLE depth_of_knowledge ADD CONSTRAINT fk__depth_of_knowledge__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+ALTER TABLE item_difficulty_cuts ADD CONSTRAINT fk__item_difficulty_cuts__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+ALTER TABLE student_group ADD CONSTRAINT fk__student_group__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+ALTER TABLE subject_claim_score ADD CONSTRAINT fk__subject_claim_score__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+
+-- Warehouse label table for holding Subject-scoped English label key/value pairs.
+-- Functionally, these should be migrated to reporting as "eng" translation values.
+CREATE TABLE subject_translation (
```
wtritch: subject_translation will only contain subject-scoped English message text.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,92 @@
...
+  subject_id TINYINT NOT NULL,
+  label_code VARCHAR(128) NOT NULL,
+  label TEXT,
+  PRIMARY KEY(subject_id, label_code),
+  CONSTRAINT fk__subject_translation__subject FOREIGN KEY (subject_id) REFERENCES subject(id)
+);
+
+-- Table for holding subject configurations in the context of an assessment type
+CREATE TABLE subject_asmt_type (
```
wtritch: This table holds the number of performance levels, sub-score performance levels, and "passing" cutoffs for subjects within the context of an assessment type.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,92 @@
...
+  sub_score_performance_level_count TINYINT,
+  sub_score_performance_level_standard_cutoff TINYINT,
+  created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY(asmt_type_id, subject_id),
+  CONSTRAINT fk__subject_asmt_type__asmt_type FOREIGN KEY (asmt_type_id) REFERENCES asmt_type(id),
+  CONSTRAINT fk__subject_asmt_type__subject FOREIGN KEY (subject_id) REFERENCES subject(id)
+);
+
+-- Create exam_target_score table for holding category target scores.
+CREATE TABLE exam_target_score (
```
wtritch: This table holds the upcoming target scores... I'm tempted to remove this since it's not technically an aspect of configurable subjects.  Thoughts @malaffoon @jtreuting @agorina ?

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,92 @@
...
+
+-- Insert data for Math: 1, ELA: 2
+-- ICA: 1, IAB: 2, SUM: 3
+INSERT INTO subject_asmt_type (asmt_type_id, subject_id, performance_level_count, performance_level_standard_cutoff, sub_score_performance_level_count, sub_score_performance_level_standard_cutoff) VALUES
+  (1, 1, 4, 3, 3, null),
+  (2, 1, 3, null, null, null),
+  (3, 1, 4, 3, 3, null),
+  (1, 2, 4, 3, 3, null),
+  (2, 2, 3, null, null, null),
+  (3, 2, 4, 3, 3, null);
```
wtritch: This inserts known data for ELA and Math into the DB.  The only values that exist in the xml that *do not* exist in the DB after flyway migration are the translation messages.

warehouse/sql/V1_2_0_3__prepare_configurable_subjects.sql
```diff
@@ -0,0 +1,142 @@
...
+  sub_score_performance_level_standard_cutoff TINYINT,
+  created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY(asmt_type_id, subject_id),
+  CONSTRAINT fk__subject_asmt_type__asmt_type FOREIGN KEY (asmt_type_id) REFERENCES asmt_type(id),
+  CONSTRAINT fk__subject_asmt_type__subject FOREIGN KEY (subject_id) REFERENCES subject(id)
+);
+
+-- Add display_order information to scorable claims.
+ALTER TABLE subject_claim_score
+  ADD COLUMN display_order TINYINT;
```
wtritch: Add a display_order column to store visual ordering information for scorable claims.

wtritch:
Closing in favor of a clean PR

PR #267, opened 2018-05-02
Remove reporting.translation.namespace column since it is no longer r…

wtritch:
…equired or respected.

I don't believe MySQL handles modifying a PRIMARY KEY very well, so I'm just creating a new table, copying values, then renaming it over the top of the old table.  Thoughts?

### DIFF
--- a/reporting/sql/V1_2_0_5__remove_translation_namespace.sql
+++ b/reporting/sql/V1_2_0_5__remove_translation_namespace.sql
@@ -3,24 +3,10 @@
 
 use ${schemaName};
 
--- Create the new translation table
-CREATE TABLE new_translation (
-  label_code VARCHAR(128),
-  language_code VARCHAR(3),
-  label text,
-  PRIMARY KEY (language_code, label_code)
-);
-
--- Migrate existing translations into new table
-INSERT INTO new_translation (label_code, language_code, label)
-  SELECT
-    label_code,
-    language_code,
-    label
-  FROM translation;
-
-DROP TABLE translation;
-RENAME TABLE new_translation TO translation;
+ALTER TABLE translation
+  DROP PRIMARY KEY,
+  ADD PRIMARY KEY(language_code, label_code),
+  DROP COLUMN namespace;
 
 -- Remove the `namespace` column from the translation staging table.
 ALTER TABLE staging_translation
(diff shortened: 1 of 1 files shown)


reporting/sql/V1_2_0_5__remove_translation_namespace.sql
```diff
@@ -0,0 +1,27 @@
...
+    language_code,
+    label
+  FROM translation;
+
+DROP TABLE translation;
+RENAME TABLE new_translation TO translation;
+
+-- Remove the `namespace` column from the translation staging table.
+ALTER TABLE staging_translation
+  DROP COLUMN namespace;
```
wtritch: I also want to drop the namespace column from the staging table.  This will percolate down into RDW_Ingest in another PR.

agorina: Have you tried doing it in one statement:
`ALTER TABLE translation DROP PRIMARY KEY, ADD PRIMARY KEY (language_code, label_code)`.

wtritch: hmmm let me give that a shot

wtritch: works like a charm, thanks @agorina !

reporting/sql/V1_2_0_5__remove_translation_namespace.sql
```diff
@@ -0,0 +1,13 @@
+-- Remove the `namespace` column from the translation table since we no longer
+-- require or respect it in the reporting application.
+
+use ${schemaName};
+
+ALTER TABLE translation
+  DROP PRIMARY KEY,
+  ADD PRIMARY KEY(language_code, label_code),
+  DROP COLUMN namespace;
```
malaffoon: i'm assuming this didn't lead to any conflicts

reporting/sql/V1_2_0_5__remove_translation_namespace.sql
```diff
@@ -0,0 +1,20 @@
...
+
+-- Rename translation table to accommodation_translation
+RENAME TABLE translation TO accommodation_translation;
```
wtritch: After talking through the future ingest process with @malaffoon and @agorina we decided to have multiple translation tables to allow warehouse and the migration process to fully manage the contents of the tables without needed to worry about namespace scoping.  When configurable subjects are introduced, we will introduce a subject_translation table to hold subject-scoped translations.

reporting/sql/V1_2_0_5__remove_translation_namespace.sql
wtritch: no sir

PR #268, opened 2018-05-08
adds first pass at user defined entity db schema and teacher student groups

esotrope:
Some things intentionally omitted for the first draft and initial review:
1. Support tables (student, school, subject and other tables). 
2. Indices and foreign keys

Wanted to leave the question open - should this db have support tables and be a full-fledged data mart with migration? should consuming services resolve support data using codes/natural ids in their respective service layers?

Let me know what you guys think.

FYI: This DB is also intended to house user_report & user_report_metadata

### DIFF
--- a/user/V1_0_0_0__ddl.sql
+++ b/user/V1_0_0_0__ddl.sql
@@ -1,0 +1,37 @@
+/*
+Initial script for the SBAC RDW Reporting Datamart 1.0.0 schema
+
+NOTES
+This schema assumes the following:
+   1. one state (aka tenant) per data warehouse
+   2. not all data elements from TRT are included, only those that are required for the current reporting
+   3. MySQL treats FK this way:
+         In the referencing table, there must be an index where the foreign key columns are listed as the first columns in the same order.
+         Such an index is created on the referencing table automatically if it does not exist.
+         This index is silently dropped later, if you create another index that can be used to enforce the foreign key constraint.
+         When restoring a DB from a back up, MySQL does not see an automatically created FK index as such and treats it as a user defined.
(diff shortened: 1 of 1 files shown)


user/V1_0_0_0__ddl.sql
```diff
@@ -0,0 +1,37 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS teacher_student_group_membership (
+  teacher_student_group_id int NOT NULL,
+  student_id int NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS teacher_student_group_subject (
+  teacher_student_group_id int NOT NULL,
+  subject_id int NOT NULL
```
esotrope: unlike `user_student_group` i broke the subject membership out into its own table to support multiple subjects (not just one or all) let me know what you guys think about this.

user/V1_0_0_0__ddl.sql
```diff
@@ -0,0 +1,37 @@
+/*
+Initial script for the SBAC RDW Reporting Datamart 1.0.0 schema
+
+NOTES
+This schema assumes the following:
+   1. one state (aka tenant) per data warehouse
+   2. not all data elements from TRT are included, only those that are required for the current reporting
```
malaffoon: 2 can probably be removed  ;)

user/V1_0_0_0__ddl.sql
malaffoon: Since this would imply a change to regular student groups i think we should keep them in sync. So, for now, i'd* leave it the same. 

* My opinions do not necessarily reflect those of my employer.

malaffoon:
I suspect we'll need an extended parking lot to hash out your questions.

PR #269, opened 2018-05-10
schema changes for unbatching group processing

malaffoon:
This is the schema change to go with code changes in ingest. There will be a coordination issue since this requires changes to schema, import and reporting all at once.


warehouse/sql/group_admin_hack.sql
```diff
@@ -1,648 +0,0 @@
-
```
malaffoon: Alla, because of the changes, none of this will work. Is it okay to delete this or do i have to figure out replacement SQL?

agorina: I should have been deleted a while ago, no need to replace it. 
Thank you for asking.

PR #270, opened 2018-05-10
adds teacher student group tables to reporting schema

esotrope:


reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
```diff
@@ -0,0 +1,26 @@
...
+  CONSTRAINT fk__teacher_student_group__school_year FOREIGN KEY (school_year) REFERENCES school_year(year)
+);
+
+CREATE TABLE IF NOT EXISTS teacher_student_group_membership (
+  teacher_student_group_id int NOT NULL,
+  student_id int NOT NULL,
+  UNIQUE INDEX idx__teacher_student_group_membership (teacher_student_group_id, student_id),
+  INDEX idx_teacher_student_group_membership__student (student_id),
+  CONSTRAINT fk__teacher_student_group_membership__student_group FOREIGN KEY (teacher_student_group_id) REFERENCES teacher_student_group(id),
+  CONSTRAINT fk__teacher_student_group_membership__student FOREIGN KEY (student_id) REFERENCES student(id)
```
esotrope: all indices and constraints were copied without question from student_group and student_group_membership tables respectively

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
```diff
@@ -0,0 +1,26 @@
+-- Adds teacher student group related tables
+
+use ${schemaName};
+
+CREATE TABLE IF NOT EXISTS teacher_student_group (
+  id int NOT NULL PRIMARY KEY,
```
agorina: Do we want this to be `AUTO_INCREMENT`?

esotrope: that would make sense yeah.

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
```diff
@@ -0,0 +1,26 @@
...
+  name varchar(255) NOT NULL,
+  school_id int NOT NULL,
+  school_year smallint NOT NULL,
+  subject_id tinyint,
+  user_login varchar(255) NOT NULL,
+  UNIQUE INDEX idx__teacher_student_group__school_name_year (school_id, name, school_year),
+  INDEX idx__teacher_student_group__subject (subject_id),
```
agorina: Please add  `INDEX idx__teacher_student_group__school_year (school_year)`.
It is needed to better control FK indexes.

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
```diff
@@ -0,0 +1,26 @@
...
+  CONSTRAINT fk__teacher_student_group__school FOREIGN KEY (school_id) REFERENCES school(id),
+  CONSTRAINT fk__teacher_student_group__subject FOREIGN KEY (subject_id) REFERENCES subject(id),
+  CONSTRAINT fk__teacher_student_group__school_year FOREIGN KEY (school_year) REFERENCES school_year(year)
+);
+
+CREATE TABLE IF NOT EXISTS teacher_student_group_membership (
+  teacher_student_group_id int NOT NULL,
+  student_id int NOT NULL,
+  UNIQUE INDEX idx__teacher_student_group_membership (teacher_student_group_id, student_id),
+  INDEX idx_teacher_student_group_membership__student (student_id),
```
agorina: Nit: it should be a double-dash between `idx` and `teacher...`

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
```diff
@@ -0,0 +1,26 @@
...
+  CONSTRAINT fk__teacher_student_group_membership__student_group FOREIGN KEY (teacher_student_group_id) REFERENCES teacher_student_group(id),
+  CONSTRAINT fk__teacher_student_group_membership__student FOREIGN KEY (student_id) REFERENCES student(id)
+);
```
agorina: Do we also need `user_teacher_student_group`?

esotrope: teachers aren't going to share groups right?

agorina: Not sure I follow your question.

How would the system know what group at teacher has access too?

esotrope: `teacher_student_group.user_login`

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
```diff
@@ -0,0 +1,27 @@
+-- Adds teacher student group related tables
+
+use ${schemaName};
+
+CREATE TABLE IF NOT EXISTS teacher_student_group (
+  id int NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  name varchar(255) NOT NULL,
```
malaffoon: nit: in general we've been avoiding 2^n-1 values and just using, for example, VARCHAR(250).

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
```diff
@@ -0,0 +1,27 @@
...
+  school_id int NOT NULL,
+  school_year smallint NOT NULL,
+  subject_id tinyint,
+  user_login varchar(255) NOT NULL,
+  UNIQUE INDEX idx__teacher_student_group__school_name_year (school_id, name, school_year),
```
agorina: Since the 'user_login` is here it may make sense to add it to the `UNIQUE INDEX`. This way different teachers could have the same name.

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
```diff
@@ -0,0 +1,27 @@
+-- Adds teacher student group related tables
+
+use ${schemaName};
+
+CREATE TABLE IF NOT EXISTS teacher_student_group (
```
malaffoon: I don't like putting `IF NOT EXISTS` in these scripts. If somebody is trying to apply this script to a system that already has this table, then something went wrong somewhere and we should fast-fail.

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
esotrope: @malaffoon be careful, i might have to take your backend dev credentials away for that

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
agorina: Since `user_login` is here we may want to add it to the `UNIQUE INDEX`.

And I am assuming that we may want to query for the teacher groups by the `user_login`, so it may need to be indexed. We can do it later when we know how we are going to query (as long as we remember to add it).

esotrope: i think our query would be `SELECT ... FROM teacher_student_group WHERE user_login = :username` right? so should it just be an index on that field?

agorina: In this case, yes - just add an index on that field.

reporting/sql/V1_2_0_6__add_teacher_student_groups.sql
malaffoon: I would prefer to use 250 but, yeah, there are a couple other exceptions that you could cite to leave it at 255. Won't lose sleep over it.

PR #272, opened 2018-05-14
Added exam_target_score.

agorina:


warehouse/sql/V1_2_0_6__exam_target_scores.sql
```diff
@@ -0,0 +1,17 @@
...
+
+CREATE TABLE exam_target_score (
+  id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  exam_id bigint NOT NULL,
+  target_id smallint NOT NULL,
+  student_relative_residual_score float,
+  student_relative_residual_score_std_err float,
+  standard_met_relative_residual_score float,
+  standard_met_relative_residual_score_std_err float,
+  performance_level tinyint,
```
agorina: Do we want to name it `performance_level` or `category` or something else?

warehouse/sql/V1_2_0_6__exam_target_scores.sql
```diff
@@ -0,0 +1,17 @@
...
+  INDEX idx__exam_claim_score__exam (exam_id),
+  UNIQUE INDEX idx__exam_target_score_exam_target (exam_id, target_id),
+  CONSTRAINT fk__exam_target_score__exam FOREIGN KEY (exam_id) REFERENCES exam(id),
+  CONSTRAINT fk__exam_target_score__target FOREIGN KEY (target_id) REFERENCES target(id)
+);
```
agorina: Here are additional columns that we have talked about and I have not added them since we are not including them into `exam_claim_score` or `exam`. I am planning on creating a JIRA for it and if we have time we can address it:
```
  ADD COLUMN item_count smallint, -- this exists in exam, but not exam_claim_score
  ADD COLUMN item_count_scored smallint, 
  ADD COLUMN raw_score int, -- what is the data type?
  ADD COLUMN raw_score_std_err float, -- what is the data type?
```

agorina: Here is what I am planning on running on AWS Dev to create dummy target scores:
```
INSERT IGNORE exam_target_score (exam_id, target_id, student_relative_residual_score, student_relative_residual_score_std_err, standard_met_relative_residual_score, standard_met_relative_residual_score_std_err, performance_level)
SELECT
  e.id AS exam_id,
  i.target_id,
  (e.scale_score +  FLOOR(rand()*10) + i.position ),
  (e.scale_score_std_err -  FLOOR(rand()) + i.position),
  (e.scale_score -  FLOOR(rand()*10) +i.position ),
  (e.scale_score_std_err +  FLOOR(rand()) - i.position),
  e.performance_level
FROM exam e
  JOIN asmt a ON a.id = e.asmt_id
  JOIN item i ON i.asmt_id = a.id
WHERE a.type_id <> 2;
```
Please provide your feedback.

warehouse/sql/V1_2_0_6__exam_target_scores.sql
malaffoon: i'm fine with performance_level but what does it refer to, the standard_met or the student_relative? Yes, i know the answer but it seems like we should name related things with related prefixes.

warehouse/sql/V1_2_0_6__exam_target_scores.sql
malaffoon: random is random; i'll send an email to Matt i guess

warehouse/sql/V1_2_0_6__exam_target_scores.sql
agorina: Do you mind sharing the answer? (because I do not know it). The requirements page I am looking at does not say it or I am failing to find it.

malaffoon: IMO, it only makes sense for the standard-met measure, just like for claims and overall. I guess, in theory, they could provide some sort of categorization of both scores but i don't think the student-relative would be below/near/above, more like worse/same/better.

PR #274, opened 2018-05-18
Added tables for target report.

agorina:
I will not merge this PR until I have SQL queries working with these tables, but would like to get feedback.

Note that I am not including staging tables yet, this will come later.

### DIFF
--- a/reporting_olap/sql/V1_0_0_0__ddl.sql
+++ b/reporting_olap/sql/V1_0_0_0__ddl.sql
@@ -290,14 +290,14 @@ CREATE TABLE asmt (
 
 CREATE TABLE target (
   id smallint NOT NULL PRIMARY KEY SORTKEY,
-  code varchar(10) NOT NULL,
+  natural_id varchar(20) NOT NULL,
   claim_code varchar(10) NOT NULL
 ) DISTSTYLE ALL;
 
 CREATE TABLE asmt_target (
   target_id int encode raw NOT NULL,
   asmt_id int encode raw NOT NULL,
-  include_into_report boolean NOT NULL,
+  include_in_report boolean NOT NULL,
   CONSTRAINT fk__asmt_target__target FOREIGN KEY(target_id) REFERENCES target(id),
   CONSTRAINT fk__asmt_target__asmt FOREIGN KEY(asmt_id) REFERENCES asmt(id)
 ) DISTSTYLE ALL;
--- a/reporting_olap/sql/V1_0_0_0__ddl.sql
+++ b/reporting_olap/sql/V1_0_0_0__ddl.sql
@@ -288,6 +288,20 @@ CREATE TABLE asmt (
   CONSTRAINT fk__asmt__school_year FOREIGN KEY(school_year) REFERENCES school_year(year)
 ) DISTSTYLE ALL COMPOUND SORTKEY (id, type_id, grade_id, subject_id);
 
+CREATE TABLE target (
+  id smallint NOT NULL PRIMARY KEY SORTKEY,
+  code varchar(10) NOT NULL,
+  claim_code varchar(10) NOT NULL
+) DISTSTYLE ALL;
+
+CREATE TABLE asmt_target (
+  target_id int encode raw NOT NULL,
+  asmt_id int encode raw NOT NULL,
+  include_into_report boolean NOT NULL,
+  CONSTRAINT fk__asmt_target__target FOREIGN KEY(target_id) REFERENCES target(id),
+  CONSTRAINT fk__asmt_target__asmt FOREIGN KEY(asmt_id) REFERENCES asmt(id)
+) DISTSTYLE ALL;
+
 CREATE TABLE asmt_active_year (
   asmt_id int NOT NULL,
   school_year smallint NOT NULL,
@@ -397,6 +411,7 @@ CREATE TABLE fact_student_iab_exam (
 
 -- Exams data for the longitudinal report.
 -- While the tables structure is similar to the other tables, the loaded data is filtered based on the different rules
+-- and includes Summative assessments only
 CREATE TABLE fact_student_exam_longitudinal (
   id bigint encode delta NOT NULL PRIMARY KEY,
   school_id integer encode raw NOT NULL,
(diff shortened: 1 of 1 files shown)


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -288,6 +288,20 @@ CREATE TABLE asmt (
   CONSTRAINT fk__asmt__school_year FOREIGN KEY(school_year) REFERENCES school_year(year)
 ) DISTSTYLE ALL COMPOUND SORTKEY (id, type_id, grade_id, subject_id);
 
+CREATE TABLE target (
+  id smallint NOT NULL PRIMARY KEY SORTKEY,
+  code varchar(10) NOT NULL,
+  claim_code varchar(10) NOT NULL
+) DISTSTYLE ALL;
```
agorina: This will be used to return target and claim codes for each report row.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -288,6 +288,20 @@ CREATE TABLE asmt (
...
+
+CREATE TABLE asmt_target (
+  target_id int encode raw NOT NULL,
+  asmt_id int encode raw NOT NULL,
+  include_into_report boolean NOT NULL,
+  CONSTRAINT fk__asmt_target__target FOREIGN KEY(target_id) REFERENCES target(id),
+  CONSTRAINT fk__asmt_target__asmt FOREIGN KEY(asmt_id) REFERENCES asmt(id)
```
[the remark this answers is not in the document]
jtreuting: I was originally thinking of a table that only include rows that needed to be excluded, but this wouldn't help with backfilling.  The reason I was thinking of it this way was that at the beginning of the year SB dev ops could run some SQL to mark which ones should be excluded.

When an assessment is loaded, I'm assuming we know all the targets and claims for it and therefore can populate this table with everything.  Is that the case?

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -476,6 +491,46 @@ CREATE TABLE fact_exam_claim_score (
...
+  exam_id bigint encode delta NOT NULL,
+  target_id int encode raw NOT NULL,
+  asmt_id int encode raw NOT NULL,
+  school_id integer encode raw NOT NULL,
+  student_id bigint encode raw NOT NULL DISTKEY,
+  grade_id smallint encode lzo NOT NULL,
+  school_year smallint encode raw NOT NULL,
+  iep smallint encode lzo NOT NULL,
+  lep smallint encode lzo NOT NULL,
+  elas_id smallint encode lzo NOT NULL,
```
jtreuting: if a state does not use ELAS would we have a value for this?

reporting_olap/sql/V1_0_0_0__ddl.sql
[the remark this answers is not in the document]
agorina: The `item` table in the warehouse looks like this:
```
CREATE TABLE IF NOT EXISTS item (
  claim_id smallint NOT NULL,
  target_id smallint NOT NULL,
  natural_id varchar(40) NOT NULL,
  asmt_id int NOT NULL,
...
```
Based on this I am assuming we can find all targets per assessment.

reporting_olap/sql/V1_0_0_0__ddl.sql
agorina: Hmmm...I do not know. I copied it from the other fact table. 

I see that it is optional in the warehouse.  I am guessing that the answer to your question maybe postponed until we have configurable optional fields, but would need @malaffoon to help me with this.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -476,6 +491,46 @@ CREATE TABLE fact_exam_claim_score (
...
+  migrant_status smallint encode lzo NOT NULL,
+  completeness_id smallint encode lzo NOT NULL,
+  administration_condition_id smallint encode lzo NOT NULL,
+  student_relative_residual_score float encode bytedict NOT NULL,
+  standard_met_relative_residual_score float encode bytedict NOT NULL,
+  completed_at timestamptz encode lzo NOT NULL,
+  migrate_id bigint encode delta NOT NULL,
+  updated timestamptz NOT NULL,
+  update_import_id bigint encode delta NOT NULL,
+  CONSTRAINT fk__fact_exam_target_score__exam FOREIGN KEY(exam_id) REFERENCES fact_student_exam(id),
```
agorina: Notice a FK to another fact table:`fact_student_exam`. Since for the target report we need to display an average scale score I am planning  on using this able and wanted to define dependency.

reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -288,6 +288,20 @@ CREATE TABLE asmt (
...
+CREATE TABLE target (
+  id smallint NOT NULL PRIMARY KEY SORTKEY,
+  code varchar(10) NOT NULL,
+  claim_code varchar(10) NOT NULL
+) DISTSTYLE ALL;
+
+CREATE TABLE asmt_target (
+  target_id int encode raw NOT NULL,
+  asmt_id int encode raw NOT NULL,
+  include_into_report boolean NOT NULL,
```
malaffoon: nit: `include_in_report`

reporting_olap/sql/V1_0_0_0__ddl.sql
malaffoon: ELAS and LEP both have to be optional.

agorina: Thank you, I will create a bug to look at it at some point.

reporting_olap/sql/V1_0_0_0__ddl.sql
malaffoon: It is a little misleading but i can't think of a better name right now.

reporting_olap/sql/V1_0_0_0__ddl.sql
agorina: https://jira.fairwaytech.com/browse/DWR-1830

reporting_olap/sql/V1_0_0_0__ddl.sql
agorina: Thanks, changing `code` to `natural_id` here.

PR #277, opened 2018-05-28
Added exam target scores.

agorina:


reporting/sql/V1_2_0_8__exam_target_scores.sql
```diff
@@ -0,0 +1,27 @@
+-- Add exam target scores
+USE ${schemaName};
+
+--TODO: change it to NOT NULL after configurable subject migrate is done?
+ALTER TABLE target ADD natural_id varchar(20);
```
agorina: @wtritch , Do I need to somehow populate this column based on the current data (can I even do it?) or do we expect that it will be loaded as part of the configurable subject changes?

reporting/sql/V1_2_0_8__exam_target_scores.sql
```diff
@@ -0,0 +1,27 @@
...
+
+CREATE TABLE asmt_target (
+  asmt_id int NOT NULL,
+  target_id smallint NOT NULL,
+  include_in_report tinyint NOT NULL,
```
agorina: I agree that in the warehouse it makes sense to keep only excluded in this table, but I was not sure about reporting. 

In Redshift we keep all of them, but we do not have `item` in the Redshift. Here we can derive the included targets from the `item`, but if feels cleaner to have it like this.

Please share your thoughts.

reporting/sql/V1_2_0_8__exam_target_scores.sql
```diff
@@ -0,0 +1,27 @@
...
+  CONSTRAINT fk__asmt_target__asmt FOREIGN KEY(asmt_id) REFERENCES asmt(id),
+  CONSTRAINT fk__asmt_target__target FOREIGN KEY(target_id) REFERENCES target(id)
+);
+
+CREATE TABLE exam_target_score (
+  id bigint NOT NULL PRIMARY KEY,
+  exam_id bigint NOT NULL,
+  target_id smallint NOT NULL,
+  student_relative_residual_score float,
+  standard_met_relative_residual_score float,
```
agorina: I am not including the corresponding std_err since we do not need them in the reporting.

reporting/sql/V1_2_0_8__exam_target_scores.sql
malaffoon: I guess it partly depends on how many targets will be excluded. And the expected lifecycle of changing those targets. 

On one hand it seems like this won't change once the assessment is created. Ever. So why bother bringing them over to reporting.

On the other hand, filtering on this flag should be performant so what's the harm in bringing them all over. If we don't, you know they'll ask for a feature to toggle the excluded targets for administrators or something.

PR #279, opened 2018-06-01
Redshift schema changes to support target migrate.

agorina:


warehouse/sql/V1_2_0_9__asmt_target_exclusion.sql
```diff
@@ -0,0 +1,15 @@
+-- Add table to hold excluded target
+USE ${schemaName};
+
+-- A configuration table to hold targets to be excluded from the 'Target Report`.
+-- This table is assumed to be manually updated.
+-- Note that it is required to trigger a 'PACKAGE' content migrate in order for this data
```
agorina: Not exactly true for Redshift since we clean/reload for now.

malaffoon: perhaps change that note to be more generic, something like `Note: this table is migrated as part of 'PACKAGE' migration.`

warehouse/sql/V1_2_0_9__asmt_target_exclusion.sql
```diff
@@ -0,0 +1,15 @@
...
+-- to be migrated into reporting data marts.
+CREATE TABLE asmt_target_exclusion (
+  asmt_id int NOT NULL,
+  target_id smallint NOT NULL,
+  PRIMARY KEY(asmt_id, target_id),
+  INDEX idx__asmt_target__target (target_id),
```
malaffoon: curious why you think we need an index on target_id? Is that just cause there is a foreign-key on it and we want to control the index creation?

agorina: Yes, it is because of the FK.  It is MySQL _feature_ that let us manage FKs better. Otherwise it is a pain to change them.

PR #280, opened 2018-06-01
Schema changes to support the target data migration to reporting.

agorina:


warehouse/sql/V1_2_0_10__trigger_code_migrate.sql
```diff
@@ -0,0 +1,6 @@
+USE ${schemaName};
+
+--TODO: NOTE this for production release unless configurable subject change this somehow
+-- Trigger migration of all the codes; the goal is to populate target/natural_id column
+INSERT INTO import(status, content, contentType, digest)
+  SELECT 1, ic.id, 'reload codes', left(uuid(), 8) from import_content ic where name = 'CODES';
```
agorina: Hmmm...this conflicts with the migrate ITs. Not sure what is the best way to handle it. 
Remove it from Flyway?

malaffoon: I think the ITs need to change so they don't conflict. The ITs should expect the system to be in a state consistent with production.

That said, i'm not sure what the upgrade process will look like for v1.2.

reporting/sql/V1_2_0_9__target_migrate.sql
```diff
@@ -0,0 +1,28 @@
...
+  id bigint NOT NULL PRIMARY KEY,
+  exam_id bigint NOT NULL,
+  target_id smallint NOT NULL,
+  student_relative_residual_score float,
+  standard_met_relative_residual_score float
+);
+
+-- left over from the translation changes
+DELETE FROM accommodation_translation
+   WHERE label_code = 'html.system-news';
```
agorina: This was the source of the broke IT mentioned below. 

@wtritch said that this is right, and I trust Will!

PR #281, opened 2018-06-02
Trigger `asmt_target` table migration for Summative assessments.

agorina:
I have a conflicting feeling if this should be included into Flyway or if this is something that we should do manually.

 It is only needed in QA because we have Summative data.

Please let me know what you think.

### DIFF
--- a/warehouse/sql/V1_2_0_11__trigger_migarte_summative_targets.sql
+++ b/warehouse/sql/V1_2_0_11__trigger_migarte_summative_targets.sql
@@ -1,0 +1,17 @@
+USE ${schemaName};
+
+--TODO: NOTE remove this file from the production release
+-- trigger migration of summative assessments to load asmt_target table in reporting data-mart
+ INSERT INTO import (status, content, contentType, digest)
+    SELECT 0, ic.id, 'force summative asmt migrate', left(uuid(), 8) from import_content ic where name = 'PACKAGE';
+
+ SELECT max(id) INTO @import_id
+    FROM IMPORT WHERE contentType = 'force summative asmt migrate';
+
+ UPDATE asmt
+    SET update_import_id = @import_id
+    WHERE type_id = 3;
+
+ UPDATE import
+    SET status = 1
+    WHERE id = @import_id;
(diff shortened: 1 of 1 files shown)


jtreuting:
i went back and forth since it won't have an effect in production right now since there aren't any summatives.  but i'd vote for manually running it in QA but couldn't probably be convinced otherwise.

warehouse/sql/V1_2_0_11__trigger_migarte_summative_targets.sql
```diff
@@ -0,0 +1,17 @@
+USE ${schemaName};
+
+--TODO: NOTE remove this file from the production release
+-- trigger migration of summative assessments to load asmt_target table in reporting data-mart
+ INSERT INTO import (status, content, contentType, digest)
+    SELECT 0, ic.id, 'force summative asmt migrate', left(uuid(), 8) from import_content ic where name = 'PACKAGE';
+
+ SELECT max(id) INTO @import_id
+    FROM IMPORT WHERE contentType = 'force summative asmt migrate';
```
malaffoon: Should/could this just use LAST_INSERT_ID instead?

malaffoon:
This script is part of RDW. It so happens that the one customer that is using it doesn't need to execute this but, a hypothetical customer using it might need to. I'd leave it.

It would be nifty if it could be made conditional; i.e. exit the script early if there are no assessments of type=3 in the system.

warehouse/sql/V1_2_0_11__trigger_migarte_summative_targets.sql
agorina: I vaguely remembered us having issues with the LAST_INSERT_ID and decided to avoid it. Probably a wrong decision here. I will change to use it.

jtreuting:
My comment was more around the fact that is a one time thing right?  When we deploy a new system this won't be needed right and then when our existing tenant wants to start using summative it won't be needed because it will have already been run.

So I'm questioning if it's a flyway thing or a one time need type script.

agorina:
I have merged it in and it will run in QA. But I am not convinced if/why we need it in production. I left a note to consider removing it. Maybe we could/should chat tomorrow?

PR #283, opened 2018-06-04
Teacher created groups DB: Remove group name constraint

agorina:


reporting/sql/V1_2_0_10__teacher_student_groups-drop-unique_index.sql
```diff
@@ -0,0 +1,8 @@
+use ${schemaName};
+
+ALTER TABLE teacher_student_group
+  DROP INDEX idx__teacher_student_group__user_login,
+  -- removes unique index
+  DROP INDEX idx__teacher_student_group__user_login_name_school_year,
+  -- re-create index without UNIQUE constraint
+  ADD INDEX idx__teacher_student_group__user_login_name_school_year (user_login, name, school_year);
```
esotrope: so this just replaces the UNIQUE index with INDEX right?
I just realized the UI doesn't send school year to the backend when looking up groups/user groups. should we be?
i think at lease for user groups we need to so that they "auto-expire" when the system year changes

agorina: The `school_year` gets taken from the configuration of the current year. I thought it was 'by design'?

PR #284, opened 2018-06-06
DWR-1571: Create SQL report on missing data

agorina:
@jtreuting  and @malaffoon - please take another look and let me know if it is OK to merge.
This addresses both https://jira.fairwaytech.com/browse/DWR-1571 and https://jira.fairwaytech.com/browse/DWR-1572.

NOTE: Item details are not included, we are still discussing the options for it.

### DIFF
--- a/warehouse/sql/missing_data_report.sql
+++ b/warehouse/sql/missing_data_report.sql
@@ -1,9 +1,14 @@
--- Should I add a reference to this or somewhere else?
--- https://github.com/SmarterApp/RDW/blob/develop/docs/Monitoring.md
---
--- Warning : This is a CPU intensive report that may take minutes to run. It is strongly advisable to run
--- this report during the maintenance window, and while the system is inactive and the exam processors
--- are paused.
+-- -----------------------------------------------------------------------------------------------------------
+-- Warning : These are CPU intensive reports that may take minutes to run. It is strongly advisable to run
+-- them report during the maintenance window, and while the system is inactive and the exam processors are paused.
+-- -----------------------------------------------------------------------------------------------------------
+
+-- -----------------------------------------------------------------------------------------------------------
+-- Summary diagnostic reports by test administration year and assessments
+-- -----------------------------------------------------------------------------------------------------------
+
+--  A summary diagnostic report indicating which the missing key data elements.
+--  To speed up its run time, this report does not include students' race or exam item data.
 SELECT
   e.school_year                                         AS test_administration_year,
   e.asmt_id                                             AS assessment_db_id,
@@ -12,7 +17,6 @@ SELECT
   count(s.last_or_surname) / count(*) * 100             AS percent_of_results_with_student_last_or_surname,
   count(s.birthday) / count(*) * 100                    AS percent_of_results_with_student_birthday,
   count(s.gender_id) / count(*) * 100                   AS percent_of_results_with_student_gender,
-  count(se.ethnicity_id) / count(*) * 100               AS percent_of_results_with_student_race,
   count(e.grade_id) / count(*) * 100                    AS percent_of_results_with_enrolled_grade,
   count(e.iep) / count(*) * 100                         AS percent_of_results_with_iep,
   count(e.lep) / count(*) * 100                         AS percent_of_results_with_lep,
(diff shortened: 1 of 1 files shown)


warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,34 @@
+-- Should I add a reference to this or somewhere else?
+-- https://github.com/SmarterApp/RDW/blob/develop/docs/Monitoring.md
```
agorina: I will remove this, but wanted to ask this question.

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,34 @@
...
+--
+-- Warning : This is a CPU intensive report that may take minutes to run. It is strongly advisable to run
+-- this report during the maintenance window, and while the system is inactive and the exam processors
+-- are paused.
```
agorina: It took almost 10 min in AWS Dev. and here is how Aurora CPU looks:
aurora cpu

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,34 @@
...
+  count(e.economic_disadvantage) / count(*) * 100       AS percent_of_results_with_economic_disadvantage,
+  count(e.migrant_status) / count(*) * 100              AS percent_of_results_with_migrant_status,
+  count(e.administration_condition_id) / count(*) * 100 AS percent_of_results_with_administration_condition,
+  -- is this `validity`?
+  count(e.completeness_id) / count(*) * 100             AS percent_of_results_with_completeness,
+  count(e.session_id) / count(*) * 100                  AS percent_of_results_with_session_id
+FROM exam e
+  JOIN student s ON s.id = e.student_id
+  JOIN asmt a ON a.id = e.asmt_id
+  JOIN (SELECT s.id, max(se.ethnicity_id) AS ethnicity_id
```
agorina: `ethnicity` takes about a half of the run time here.

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,34 @@
...
+          FROM student s LEFT JOIN student_ethnicity se ON s.id = se.student_id
+          GROUP BY s.id
+       ) se ON se.id = e.student_id
+-- optionally include : WHERE e.deleted = 0
+GROUP BY e.asmt_id, e.school_year;
```
agorina: I am still testing it but I wanted to share this to start the discussion about `Percent of results with item details`. This is the field that is missing in this report. I have tried different versions of the SQL, but it takes forever (well I have waited an hour+ and gave up). What exactly is needed for this requirement. Can we be _creative_ and capture an item counts as a data element in the exam table when we are inserting items?

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,34 @@
...
+  count(s.gender_id) / count(*) * 100                   AS percent_of_results_with_student_gender,
+  count(se.ethnicity_id) / count(*) * 100               AS percent_of_results_with_student_race,
+  count(e.grade_id) / count(*) * 100                    AS percent_of_results_with_enrolled_grade,
+  count(e.iep) / count(*) * 100                         AS percent_of_results_with_iep,
+  count(e.lep) / count(*) * 100                         AS percent_of_results_with_lep,
+  count(e.section504) / count(*) * 100                  AS percent_of_results_with_section504,
+  count(e.economic_disadvantage) / count(*) * 100       AS percent_of_results_with_economic_disadvantage,
+  count(e.migrant_status) / count(*) * 100              AS percent_of_results_with_migrant_status,
+  count(e.administration_condition_id) / count(*) * 100 AS percent_of_results_with_administration_condition,
+  -- is this `validity`?
```
agorina: @jtreuting , is this right for what is called as `Percent of results with validity`?

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,34 @@
...
+--
+-- Warning : This is a CPU intensive report that may take minutes to run. It is strongly advisable to run
+-- this report during the maintenance window, and while the system is inactive and the exam processors
+-- are paused.
+SELECT
+  e.school_year                                         AS test_administration_year,
+  e.asmt_id                                             AS assessment_db_id,
+  a.natural_id                                          AS asessment_natural_id,
+  count(*)                                              AS total_results,
+  count(s.last_or_surname) / count(*) * 100             AS percent_of_results_with_student_last_or_surname,
```
agorina: From mySQL tutorial:
_The COUNT(\*) function returns the number of rows in a result set returned by a SELECT statement. The COUNT(expression) returns the number of rows that do not contain NULL values._

warehouse/sql/missing_data_report.sql
jtreuting: administration condition covers both Valid/Invalid and Standardized/Non-standardized right?  so yeah i think if it's empty it would cover both scenarios where summative or interim values are missing for this status type concept

warehouse/sql/missing_data_report.sql
jtreuting: we could do that but it seems like this value is probably not *that* important to warrant capturing it during the normal flow.  but that is a gut reaction.

could we put the logic in the SQL but commented out and say that they can uncomment to run but it should be done very carefully?

agorina: Let's chat once the UAT dust settles?

warehouse/sql/missing_data_report.sql
malaffoon: perhaps remove from here and have a separate query for ethnicity?

warehouse/sql/missing_data_report.sql
malaffoon: Yes, makes sense to reference it somewhere in RDW. Monitoring.md makes sense. I'm wondering if, by the end of the project, we should move these kinds of scripts into RDW. It's nice having them here so it is easier to maintain them as changes are made but they don't really belong here.

agorina: Whoever comes after us may want to have these scrips in here for the same reason we do now. Just a thought.

warehouse/sql/missing_data_report.sql
agorina: I was thinking that too, but convinced myself otherwise. Here is my chain of thoughts: if somebody with the SQL background needs to run this query, this gives a good start and they could adjust it if needed. If somebody without SQL background is asked to run it and need to produce the report as defined, they would want it in one SQL.

But I am open to break it into two? @jtreuting ?

warehouse/sql/missing_data_report.sql
agorina: thank you

warehouse/sql/missing_data_report.sql
jtreuting: i think we are open to break it into 2.  the benefit being you could do your basic analysis quicker and then decide to dig into ethnicities based on the results of the first part.  i think that works.

warehouse/sql/missing_data_report.sql
agorina: https://github.com/SmarterApp/RDW/pull/33

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,104 @@
...
+  e.school_year                                         AS test_administration_year,
+  e.asmt_id                                             AS assessment_db_id,
+  a.natural_id                                          AS asessment_natural_id,
+  count(*)                                              AS total_results,
+  count(se.ethnicity_id) / count(*) * 100               AS percent_of_results_with_student_race
+FROM exam e
+  JOIN asmt a ON a.id = e.asmt_id
+  JOIN (SELECT s.id, max(se.ethnicity_id) AS ethnicity_id FROM student s LEFT JOIN student_ethnicity se ON s.id = se.student_id GROUP BY s.id) se ON se.id = e.student_id
+-- optionally include : WHERE e.deleted = 0
+GROUP BY e.asmt_id, e.school_year;
```
agorina: Run time 4 min.

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,104 @@
...
+  count(e.economic_disadvantage) / count(*) * 100       AS percent_of_results_with_economic_disadvantage,
+  count(e.migrant_status) / count(*) * 100              AS percent_of_results_with_migrant_status,
+  count(e.administration_condition_id) / count(*) * 100 AS percent_of_results_with_administration_condition,
+  count(e.completeness_id) / count(*) * 100             AS percent_of_results_with_completeness,
+  count(e.session_id) / count(*) * 100                  AS percent_of_results_with_session_id
+FROM exam e
+  JOIN student s ON s.id = e.student_id
+  JOIN asmt a ON a.id = e.asmt_id
+-- optionally include : WHERE e.deleted = 0
+GROUP BY e.asmt_id, e.school_year;
```
agorina: Run time 4 min.

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,104 @@
...
+  count(e.completeness_id) / count(*) * 100             AS percent_of_results_with_completeness,
+  count(e.session_id) / count(*) * 100                  AS percent_of_results_with_session_id
+FROM exam e
+  JOIN student s ON s.id = e.student_id
+  JOIN asmt a ON a.id = e.asmt_id
+  JOIN school sch ON sch.id = e.school_id
+  JOIN district d ON d.id = sch.district_id
+-- optionally include : WHERE d.natural_id = 'put district id here'
+-- optionally include : AND e.deleted = 0
+GROUP BY e.asmt_id, e.school_year, e.school_id;
```
agorina: Run time 11 min.

warehouse/sql/missing_data_report.sql
```diff
@@ -0,0 +1,104 @@
...
+  count(*)                                AS total_results,
+  count(se.ethnicity_id) / count(*) * 100 AS percent_of_results_with_student_race
+FROM exam e
+  JOIN asmt a ON a.id = e.asmt_id
+  JOIN (SELECT s.id, max(se.ethnicity_id) AS ethnicity_id FROM student s LEFT JOIN student_ethnicity se ON s.id = se.student_id GROUP BY s.id) se ON se.id = e.student_id
+  JOIN school sch ON sch.id = e.school_id
+  JOIN district d ON d.id = sch.district_id
+-- optionally include : WHERE d.natural_id = 'put district id here'
+-- optionally include : AND e.deleted = 0
+GROUP BY e.asmt_id, e.school_year, e.school_id;
```
agorina: Run time 11 min.

agorina:
@jtreuting , I am merging but I will address any comments you may have.

PR #286, opened 2018-06-08
remove unused fields from exam_target_score table

malaffoon:
There are no code changes necessary to go along with this.

### DIFF
--- a/warehouse/sql/V1_2_0_13__remove_target_std_err.sql
+++ b/warehouse/sql/V1_2_0_13__remove_target_std_err.sql
@@ -1,0 +1,33 @@
+-- Drop std_err columns from target score table, they will never be populated
+
+USE ${schemaName};
+
+ALTER TABLE exam_target_score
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_2_0_13__remove_target_std_err.sql
```diff
@@ -0,0 +1,33 @@
...
+
+USE ${schemaName};
+
+ALTER TABLE exam_target_score
+    DROP COLUMN student_relative_residual_score_std_err,
+    DROP COLUMN standard_met_relative_residual_score_std_err;
+
+ALTER TABLE audit_exam_target_score
+    DROP COLUMN student_relative_residual_score_std_err,
+    DROP COLUMN standard_met_relative_residual_score_std_err;
```
malaffoon: Hmm, does it *really* matter if i do this before i change the trigger?

agorina: I do not know. As long as it runs :)

PR #289, opened 2018-06-15
Make student related data elements optional

agorina:
I am breaking it into baby step. This covers all student related attributes.


malaffoon:
Just realized this will require changes to the exam audit table as well

PR #292, opened 2018-06-16
Add index to support migrate.

agorina:


warehouse/sql/V1_2_0_17__exam_index.sql
```diff
@@ -0,0 +1,8 @@
+-- add indexes to support migrate
+use ${schemaName};
+
+ALTER TABLE exam
+ ADD INDEX idx__exam__type_deleted_created_and_scores (type_id, deleted, created, scale_score, scale_score_std_err, performance_level);
+
+ALTER TABLE exam
+ ADD INDEX idx__exam__type_deleted_updated_and_scores (type_id, deleted, updated, scale_score, scale_score_std_err, performance_level);
```
agorina: As noted in the slack msg, this is needed for the migrate (at least the OLAP migrate):
(the error points to the time the indexes were created)
run-time

There are two more compound indexes on this table:
```
 create index idx__exam__type_deleted_student_score_and_more
	on exam (type_id, deleted, student_id, scale_score, scale_score_std_err, performance_level)
;

create index idx__exam__student_school_year_asmt
	on exam (student_id, school_year, asmt_id)
;
```

And even more indexes to support FK:
```
create index fk__exam__school_year
	on exam (school_year);

create index idx__exam__asmt
	on exam (asmt_id);

create index idx__exam__created
	on exam (created);

create index idx__exam__import
	on exam (import_id);

create index idx__exam__school
	on exam (school_id);

create index idx__exam__student
	on exam (student_id);

create index idx__exam__updated
	on exam (updated);

create index idx__exam__update_import
	on exam (update_import_id);
```
Adding two more indexes on a table of this size without analyzing the impact makes me a little uncomfortable. 

Ideally it would be nice to load and migrate to Aurora more exams to see if this has any significant impact. And if some of the existing indexes cold be dropped.

Thoughts?

agorina: Pplease remind me to delete the indexes that I create in dev. once we are done with this. (they are named with `...alla..` in case you see them and wonder what it is).

agorina: And one more observation: `COPY` queries seems to be missing from the `slow_log`. Just something to be aware since we use them in the app in a few places (here and in the back-end CSV export).

agorina:
Since we need to re-migrate the data in production I feel that this needs to be merged, but I will bring it up for a discussion tomorrow.

PR #295, opened 2018-06-19
Initial schema changes for reporting to support configurable subjects.

wtritch:
This PR modifies the reporting tables to support configurable subjects.

See the comments at the top of the file for the list of changes.


reporting/sql/V1_2_1_1__configurable_subject_schema.sql
```diff
@@ -0,0 +1,157 @@
...
+);
+
+-- Insert data for Math: 1, ELA: 2
+-- ICA: 1, IAB: 2, SUM: 3
+INSERT INTO subject_asmt_type (asmt_type_id, subject_id, performance_level_count, performance_level_standard_cutoff, claim_score_performance_level_count) VALUES
+  (1, 1, 4, 3, 3),
+  (2, 1, 3, null, null),
+  (3, 1, 4, 3, 3),
+  (1, 2, 4, 3, 3),
+  (2, 2, 3, null, null),
```
esotrope: why null instead of 0?

reporting/sql/V1_2_1_1__configurable_subject_schema.sql
```diff
@@ -0,0 +1,157 @@
...
+
+-- Make name nullable for subject_claim_score
+-- and add display_order column
+ALTER TABLE subject_claim_score
+  ADD COLUMN display_order TINYINT,
+  MODIFY COLUMN name VARCHAR(250) DEFAULT NULL;
+CREATE TABLE staging_subject_claim_score (
+  id tinyint NOT NULL,
+  subject_id tinyint NOT NULL,
+  asmt_type_id tinyint NOT NULL,
```
esotrope: no fks?

reporting/sql/V1_2_1_1__configurable_subject_schema.sql
```diff
@@ -0,0 +1,157 @@
...
+-- Make display text columns nullable in preparation for removal
+-- Create a subject_translation table to hold subject-scoped display text
+
+USE ${schemaName};
+
+-- Alter subject table to hold import references
+-- TODO should we initialize import_id and update_import_id to -1 here and in warehouse?
+ALTER TABLE subject
+  ADD COLUMN created TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  ADD COLUMN updated TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
agorina: Nit: Reporting tables have just `updated` and `update_import_id`, but not `created` or `import_id`.

And you are missing `migrate_id`. We keep it for the troubleshooting/traceability purpose.

wtritch: ah, I was basing this off of the `exam` table.  Will remove `created` and `import_id` and add `migrate_id`

reporting/sql/V1_2_1_1__configurable_subject_schema.sql
```diff
@@ -0,0 +1,215 @@
...
+-- Create a staging_exam_claim_score_mapping table to support migration
+-- Create a staging_subject_claim_score to support migration
+
+USE ${schemaName};
+
+-- Drop FK references on subject.id for modification
+ALTER TABLE asmt
+  DROP FOREIGN KEY fk__asmt__subject,
+  MODIFY COLUMN subject_id SMALLINT NOT NULL;
+ALTER TABLE staging_asmt MODIFY COLUMN subject_id SMALLINT NOT NULL;
```
wtritch: These are all "referencing table" modifications to do the foreign key drop/re-add dance for changing `subject.id` from a TINYINT to a SMALLINT.  I'm also updating the associated staging tables (where they exist.)

reporting/sql/V1_2_1_1__configurable_subject_schema.sql
```diff
@@ -0,0 +1,215 @@
...
+  MODIFY COLUMN code VARCHAR(10) DEFAULT NULL,
+  MODIFY COLUMN description VARCHAR(500) DEFAULT NULL;
+
+-- Make description nullable for depths of knowledge
+ALTER TABLE item
+  DROP FOREIGN KEY fk__item__dok,
+  MODIFY COLUMN dok_id SMALLINT;
+ALTER TABLE staging_item MODIFY COLUMN dok_id SMALLINT;
+ALTER TABLE depth_of_knowledge
+  MODIFY COLUMN id SMALLINT NOT NULL,
```
wtritch: This is similar to the `subject.id` changes above.

reporting/sql/V1_2_1_1__configurable_subject_schema.sql
wtritch: Because for performance_level_standard_cutoff a cutoff of 0 implies that there's just one group, while null implies there is no grouping.
And for claim_score_performance_level_count it felt more like an "N/A" than a "no performance levels."
Although in all honesty, I could be convinced either way because this feels like an academic question. =)

reporting/sql/V1_2_1_1__configurable_subject_schema.sql
wtritch: Not on staging tables

reporting/sql/V1_2_1_1__configurable_subject_schema.sql
```diff
@@ -0,0 +1,215 @@
...
+  (1, 2, 4, 3, 3),
+  (2, 2, 3, null, null),
+  (3, 2, 4, 3, 3);
+
+-- Make name nullable for subject_claim_score
+-- and add display_order column
+ALTER TABLE exam_claim_score_mapping
+  DROP FOREIGN KEY fk__exam_claim_score_mapping__subject_claim_score,
+  MODIFY COLUMN subject_claim_score_id SMALLINT NOT NULL;
+CREATE TABLE staging_exam_claim_score_mapping (
```
agorina: I am not convinced we need this staging table. 
We should be able to define `exam_claim_score_mapping` from `subject_claim_score` by using its `id` and `display_order`, I think.

wtritch: Hm... I think that's probably how we're going to do it, but if that happens in the migration phase, don't we need a place to put it?  Or should reporting be calculating that at view time?

agorina: I think the migrate should load this table directly as the first step. Then migrate could use it to pivot the exams' claim scores and reporting could use it to relate claims to pivoted columns.

PR #297, opened 2018-06-21
More changes to support configurable subjects.

agorina:
I am not going to merge it until the migrate is done and I verified that the rest of the system works with it, but wanted to get feedback.

### DIFF
--- a/reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
+++ b/reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
@@ -22,7 +22,10 @@ ALTER TABLE common_core_standard ADD CONSTRAINT fk__common_core_standard__subjec
 ALTER TABLE subject_claim_score
     DROP FOREIGN KEY fk__subject_claim_score__subject,
     DROP INDEX idx__subject_claim_score__subject,
+    -- add a column to support claims pivoting during migrate into reporting
+    ADD COLUMN data_order TINYINT,
     ADD UNIQUE INDEX idx__subject_claim_score__subject_asmt_code(subject_id, asmt_type_id, code);
+ALTER TABLE subject_claim_score ADD UNIQUE INDEX idx__subject_claim_score__subject_asmt_data_order(subject_id, asmt_type_id, data_order);
 ALTER TABLE subject_claim_score ADD CONSTRAINT fk__subject_claim_score__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
 
 ALTER TABLE target
@@ -34,10 +37,7 @@ ALTER TABLE target ADD CONSTRAINT fk__target__claim FOREIGN KEY (claim_id) REFER
 ALTER TABLE staging_subject_claim_score
   ADD COLUMN name VARCHAR(250) DEFAULT NULL;
 
--- add a column to support claims pivoting during migrate into reporting
-ALTER TABLE subject_claim_score
- ADD COLUMN data_order TINYINT;
-
+-- load data for the existing scores
 UPDATE subject_claim_score SET data_order = 1 WHERE code = '1' AND subject_id = 1;
 UPDATE subject_claim_score SET data_order = 2 WHERE code = 'SOCK_2' AND subject_id = 1;
 UPDATE subject_claim_score SET data_order = 3 WHERE code = '3' AND subject_id = 1;
--- a/reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
+++ b/reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
@@ -31,10 +31,31 @@ ALTER TABLE target
     ADD UNIQUE INDEX idx__target__claim_natural_id(claim_id, natural_id);
 ALTER TABLE target ADD CONSTRAINT fk__target__claim FOREIGN KEY (claim_id) REFERENCES claim(id);
 
+ALTER TABLE staging_subject_claim_score
+  ADD COLUMN name VARCHAR(250) DEFAULT NULL;
+
+-- add a column to support claims pivoting during migrate into reporting
+ALTER TABLE subject_claim_score
+ ADD COLUMN data_order TINYINT;
+
+UPDATE subject_claim_score SET data_order = 1 WHERE code = '1' AND subject_id = 1;
+UPDATE subject_claim_score SET data_order = 2 WHERE code = 'SOCK_2' AND subject_id = 1;
+UPDATE subject_claim_score SET data_order = 3 WHERE code = '3' AND subject_id = 1;
+UPDATE subject_claim_score SET data_order = 1 WHERE code = 'SOCK_R' AND subject_id = 2;
+UPDATE subject_claim_score SET data_order = 2 WHERE code = 'SOCK_LS' AND subject_id = 2;
+UPDATE subject_claim_score SET data_order = 3 WHERE code = '2-W' AND subject_id = 2;
+UPDATE subject_claim_score SET data_order = 4 WHERE code = '4-CR' AND subject_id = 2;
+
+ALTER TABLE subject_claim_score
+ MODIFY COLUMN data_order TINYINT NOT NULL;
+
+ALTER TABLE staging_subject_claim_score
+ ADD COLUMN data_order TINYINT NOT NULL;
+
 -- we can now use `subject_claim_score` table with the id and `display_order`
 DROP TABLE exam_claim_score_mapping;
 
--- there is not used in the reporting
+-- these are not used in reporting
 DROP TABLE item_trait_score;
 DROP TABLE staging_item_trait_score;
 
@@ -58,7 +79,7 @@ ALTER TABLE subject
     MODIFY COLUMN update_import_id BIGINT NOT NULL,
     MODIFY COLUMN migrate_id BIGINT NOT NULL;
 
--- TODO:
+-- TODO:clean up
 --ALTER TABLE depth_of_knowledge
 --    DROP COLUMN description;
 --ALTER TABLE claim
@@ -69,5 +90,18 @@ ALTER TABLE subject
 --ALTER TABLE subject_claim_score
 --    DROP COLUMN name;
 --ALTER TABLE target
+--    DROP COLUMN code,
+--    DROP COLUMN description;
+
+--ALTER TABLE staging_depth_of_knowledge
+--    DROP COLUMN description;
+--ALTER TABLE staging_claim
+--    DROP COLUMN name,
+--    DROP COLUMN description;
+--ALTER TABLE staging_common_core_standard
+--    DROP COLUMN description;
+--ALTER TABLE staging_subject_claim_score
+--    DROP COLUMN name;
+--ALTER TABLE staging_target
 --    DROP COLUMN code,
 --    DROP COLUMN description;
--- a/reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
+++ b/reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
@@ -10,6 +10,7 @@ ALTER TABLE depth_of_knowledge ADD CONSTRAINT fk__depth_of_knowledge__subject FO
 
 ALTER TABLE claim
     DROP FOREIGN KEY fk__claim__subject,
+    DROP INDEX idx__claim__subject,
     ADD UNIQUE INDEX idx__claim__subject_code(subject_id, code);
 ALTER TABLE claim ADD CONSTRAINT fk__claim__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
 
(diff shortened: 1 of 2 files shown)


reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,73 @@
...
+DROP TABLE exam_claim_score_mapping;
+
+-- there is not used in the reporting
+DROP TABLE item_trait_score;
+DROP TABLE staging_item_trait_score;
+
+-- add up to the max supported claims
+ALTER TABLE asmt
+  ADD COLUMN claim5_score_code varchar(10),
+  ADD COLUMN claim6_score_code varchar(10);
```
agorina: `asmt` also has `cut_point_1`,`cut_point_2`,`cut_point_3`. Do I need to expand them somehow?

Anything else I am missing?

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,73 @@
...
+  ADD COLUMN claim6_category tinyint;
+
+-- warehouse schema has a change that will force subject re-migration and will update `update_import_id`
+UPDATE subject SET update_import_id = -1, migrate_id = -1;
+
+ALTER TABLE subject
+    MODIFY COLUMN update_import_id BIGINT NOT NULL,
+    MODIFY COLUMN migrate_id BIGINT NOT NULL;
+
+-- TODO:
```
agorina: Just saving it for the next script once the reporting changes are done and we can clean up.

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,73 @@
...
+ALTER TABLE subject_claim_score ADD CONSTRAINT fk__subject_claim_score__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+
+ALTER TABLE target
+    DROP FOREIGN KEY fk__target__claim,
+    DROP INDEX idx__target__claim,
+    ADD UNIQUE INDEX idx__target__claim_natural_id(claim_id, natural_id);
+ALTER TABLE target ADD CONSTRAINT fk__target__claim FOREIGN KEY (claim_id) REFERENCES claim(id);
+
+-- we can now use `subject_claim_score` table with the id and `display_order`
+DROP TABLE exam_claim_score_mapping;
```
agorina: This concerns me a little bit.I am re-purposing the `display_order` order to match it to the pivoted columns. In other words, a claim with the display order =1 will be in the `exam`'s`claim1_scale_score` column.

Do we allow changes to the display order once the subject is ingested? 
Is this a bad idea to do it this way?

jtreuting: that's a tough one, i think it deserves a discussion.

might not be relevant or important but it reminded me.  right now the column order for ELA claims does not match the display order.  it was something i had to address with the CSV export when we reordered the claims to match the logical order they used.

agorina: Let's chat tomorrow then, thank you!

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,73 @@
...
+
+-- there is not used in the reporting
```
malaffoon: grammer nazi nit: `these are not used in reporting`

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
malaffoon: Those correspond to max performance levels, right?

wtritch: yeah... we'll need 5 if we support 6 performance levels...

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
wtritch: Yeah... we don't enforce that display order is sequential or starts at 1. We would need to specify and validate that if we want to use it this way.

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,73 @@
...
+  ADD COLUMN claim5_scale_score_std_err float,
+  ADD COLUMN claim5_category tinyint,
+  ADD COLUMN claim6_scale_score smallint,
+  ADD COLUMN claim6_scale_score_std_err float,
+  ADD COLUMN claim6_category tinyint;
+
+-- warehouse schema has a change that will force subject re-migration and will update `update_import_id`
+UPDATE subject SET update_import_id = -1, migrate_id = -1;
+
+ALTER TABLE subject
```
wtritch: Weren't you also going to modify the id column to a smallint?  And a few other tables too?

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,73 @@
...
+--ALTER TABLE claim
+--    DROP COLUMN name,
+--    DROP COLUMN description;
+--ALTER TABLE common_core_standard
+--    DROP COLUMN description;
+--ALTER TABLE subject_claim_score
+--    DROP COLUMN name;
+--ALTER TABLE target
+--    DROP COLUMN code,
+--    DROP COLUMN description;
```
wtritch: Maybe add a TODO to drop these from the reporting (and staging) tables as well?

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
agorina: @wtritch , does it mean that we need to update `warehouse`,`asmt_score` to have up to 5 cut_points?  Would this require an XSD change and subject ingest change, or asmt ingest change? (I may need your help with it if this is the case).

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
agorina: Good point, I will add staging. The rest should be covered, I think.

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
agorina: This was done in `V1_2_1_0__configurable_subject.sql`.

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
agorina: Talked offline. Will change the schema and asmt ingest.

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,107 @@
...
+ALTER TABLE subject_claim_score
+ ADD COLUMN data_order TINYINT;
+
+UPDATE subject_claim_score SET data_order = 1 WHERE code = '1' AND subject_id = 1;
+UPDATE subject_claim_score SET data_order = 2 WHERE code = 'SOCK_2' AND subject_id = 1;
+UPDATE subject_claim_score SET data_order = 3 WHERE code = '3' AND subject_id = 1;
+UPDATE subject_claim_score SET data_order = 1 WHERE code = 'SOCK_R' AND subject_id = 2;
+UPDATE subject_claim_score SET data_order = 2 WHERE code = 'SOCK_LS' AND subject_id = 2;
+UPDATE subject_claim_score SET data_order = 3 WHERE code = '2-W' AND subject_id = 2;
+UPDATE subject_claim_score SET data_order = 4 WHERE code = '4-CR' AND subject_id = 2;
```
wtritch: I don't think we need data_order in reporting?  This is covered by asmt.claim1_code, asmt.claim2_code, etc?

agorina: We need it in the reporting to support migrate.

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,107 @@
...
+-- To avoid redundant indexes, first dropped FK and Indexes that supports them, and re-created FK afterwards
+ALTER TABLE depth_of_knowledge
+    DROP FOREIGN KEY fk__depth_of_knowledge__subject,
+    DROP INDEX idx__depth_of_knowledge__subject,
+    ADD UNIQUE INDEX idx__depth_of_knowledge__subject_level(subject_id, level);
+ALTER TABLE depth_of_knowledge ADD CONSTRAINT fk__depth_of_knowledge__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+
+ALTER TABLE claim
+    DROP FOREIGN KEY fk__claim__subject,
+    ADD UNIQUE INDEX idx__claim__subject_code(subject_id, code);
```
malaffoon: this one looks different than the others, it isn't dropping an index after the FK

warehouse/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,78 @@
...
+
+ALTER TABLE item_difficulty_cuts
+    DROP FOREIGN KEY fk__item_difficulty_cuts__subject,
+    DROP INDEX idx__item_difficulty_cuts__subject,
+    ADD UNIQUE INDEX idx__tem_difficulty_cuts__subject_grade_diff(subject_id, grade_id);
+ALTER TABLE item_difficulty_cuts ADD CONSTRAINT fk__item_difficulty_cuts__subject FOREIGN KEY (subject_id) REFERENCES subject(id);
+
+ALTER TABLE claim
+    DROP FOREIGN KEY fk__claim__subject,
+    ADD UNIQUE INDEX idx__claim__subject_code(subject_id, code);
```
malaffoon: ditto

warehouse/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,78 @@
...
+UPDATE subject_claim_score SET data_order = 3 WHERE code = '2-W' AND subject_id = 2;
+UPDATE subject_claim_score SET data_order = 4 WHERE code = '4-CR' AND subject_id = 2;
+
+-- Create SUBJECT import content and trigger re-import to synch-up the data and import ids
+INSERT INTO import_content (id, name) VALUES (8, 'SUBJECT');
+
+INSERT INTO import (status, content, contentType, digest) VALUES (0, 8, 'config subject support', 'config subject support v1.2.1');
+SELECT LAST_INSERT_ID() INTO @import_id;
+UPDATE subject SET import_id = @import_id, update_import_id = @import_id;
+
```
malaffoon: shouldn't something be setting the import status to 1?

warehouse/sql/V1_2_1_1__configurable_subject_tighten_up.sql
```diff
@@ -0,0 +1,78 @@
...
+ALTER TABLE subject
+    MODIFY COLUMN import_id BIGINT NOT NULL,
+    MODIFY COLUMN update_import_id BIGINT NOT NULL;
```
malaffoon: seems odd to modify the table *after* triggering the import, i'd prefer to move this up

reporting/sql/V1_2_1_1__configurable_subject_tighten_up.sql
agorina: thank you! I missed it, adding it.

warehouse/sql/V1_2_1_1__configurable_subject_tighten_up.sql
agorina: thanks again.

warehouse/sql/V1_2_1_1__configurable_subject_tighten_up.sql
agorina: Yes, missed it again. Thank you!

warehouse/sql/V1_2_1_1__configurable_subject_tighten_up.sql
agorina: Good point, but I am not sure how to do it? The migrate starts as soon as we insert the import. Any suggestions?

malaffoon: well, the system shouldn't be running when these are applied

jtreuting: wouldn't it be moving it up to line 50 before inserting into the import table

agorina: Talked offline. I cannot do it because I need to populated the columns with the import ids  before making them not nullable. And the import id is not created before we insert into the `import` table.

PR #299, opened 2018-06-25
Added up to 5 cut points to the assessment.

agorina:


warehouse/sql/V1_2_1_2__asmt_cut_points.sql
```diff
@@ -0,0 +1,10 @@
+use ${schemaName};
+
+-- add up to 5 cut points
+ALTER TABLE asmt_score
+-- allow nulls
+ MODIFY COLUMN cut_point_2 float,
+ MODIFY COLUMN min_score float,
+ MODIFY COLUMN max_score float,
```
agorina: This supports `performanceLevelCount = 0`. Do we want/need it? Or should we restrict to some min number?

PR #300, opened 2018-06-25
Relax uniqueness constraints on subject_claim_score to allow for upda…

wtritch:
…ting data_order.

### DIFF
--- a/warehouse/sql/V1_2_1_3__relax_uniqueness.sql
+++ b/warehouse/sql/V1_2_1_3__relax_uniqueness.sql
@@ -1,0 +1,6 @@
+-- Relax UNIQUE requirements to allow for re-ordering scorable claim data orders
+
+use ${schemaName};
+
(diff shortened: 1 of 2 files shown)


warehouse/sql/V1_2_1_3__relax_uniqueness.sql
```diff
@@ -0,0 +1,6 @@
+-- Relax UNIQUE requirements to allow for re-ordering scorable claim data orders
+
+use ${schemaName};
+
+ALTER TABLE subject_claim_score
+  DROP INDEX idx__subject_claim_score__subject_asmt_data_order;
```
agorina: You may want to do that in the reporting too. Not sure if migrate scripts will handle the update correctly.

wtritch: Good call, done

PR #301, opened 2018-06-25
fix some obvious discrepancies for validating the reporting data mart

malaffoon:
This is a quick fix for the obvious issues. It doesn't (yet) add any new validation.

### DIFF
--- a/validation/sql/total-iab-by-asmt-schoolyear-condition-complete/reporting.sql
+++ b/validation/sql/total-iab-by-asmt-schoolyear-condition-complete/reporting.sql
@@ -3,7 +3,7 @@ SELECT
     e.asmt_id,
     e.school_year,
     e.administration_condition_code,
-    CASE WHEN e.completeness_code = 'Complete' THEN 'TRUE' ELSE 'FALSE' END
+    e.completeness_code
   FROM exam e
   WHERE e.type_id = 2
   GROUP BY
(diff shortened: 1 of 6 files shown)


validation/sql/total-iab-by-asmt-schoolyear-condition-complete/reporting.sql
```diff
@@ -3,7 +3,7 @@ SELECT
     e.asmt_id,
     e.school_year,
     e.administration_condition_code,
-    CASE WHEN e.completeness_code = 'Complete' THEN 'TRUE' ELSE 'FALSE' END
+    e.completeness_code
```
malaffoon: Basic idea: let reporting just query codes while the warehouse will join to the reference table to look up the code.

validation/sql/total-iab-scores/warehouse.sql
```diff
@@ -1,6 +1,6 @@
 SELECT
     sum(round(scale_score)),
-    sum(scale_score_std_err),
+    round(sum(scale_score_std_err), 1),
```
malaffoon: To avoid emitting the meaningless diff on std-err, this rounds to the first decimal. 

NOTE: this is very different than the `round` for the scale-score which is there because we round values before sending them to reporting data mart. That's why this one is round-the-sum while the other is sum-the-rounded-values.

PR #302, opened 2018-06-26
Add support to null values.

agorina:
There is an outstanding JIRA to adjust the migrate to work with the NULL values, but this is a patch to get the QA working.


agorina:
Merging to test it in QA, but will address comments as needed.

PR #305, opened 2018-06-27
Relax NOT NULL constraints on item.target_code in reporting schema.

wtritch:


reporting/sql/V1_2_1_4__relax_constraints.sql
```diff
@@ -0,0 +1,6 @@
+-- Relax not null constraints on nullable columns.
+
+use ${schemaName};
+
+ALTER TABLE item
+  MODIFY COLUMN target_code VARCHAR(10) DEFAULT NULL;
```
agorina: Why not drop it?

wtritch: Because then i’d Have to modify migration >_<

agorina: Could you please create a JIRA for it?

PR #308, opened 2018-07-06
consolidate flyway scripts for 1.2.1

malaffoon:
I've tested that they run fine but i still need to test loading data, etc. against the result. However, i'm going to be off for a couple hours so i thought i'd get eyes on it.


reporting/sql/V1_2_1_0__update.sql
```diff
@@ -0,0 +1,259 @@
...
+  CONSTRAINT fk__subject_translation__subject FOREIGN KEY (subject_id) REFERENCES subject(id)
+);
+CREATE TABLE staging_subject_translation (
+  subject_id SMALLINT NOT NULL,
+  label_code VARCHAR(128) NOT NULL,
+  label TEXT,
+  PRIMARY KEY(subject_id, label_code)
+);
+
+-- Make code and description nullable for targets
```
wtritch: nit, we're dropping those columns rather than making them nullable now =)

reporting/sql/V1_2_1_0__update.sql
```diff
@@ -0,0 +1,259 @@
...
+-- ICA: 1, IAB: 2, SUM: 3
+INSERT INTO subject_asmt_type (asmt_type_id, subject_id, performance_level_count, performance_level_standard_cutoff, claim_score_performance_level_count) VALUES
+  (1, 1, 4, 3, 3),
+  (2, 1, 3, null, null),
+  (3, 1, 4, 3, 3),
+  (1, 2, 4, 3, 3),
+  (2, 2, 3, null, null),
+  (3, 2, 4, 3, 3);
+
+-- we can now use `subject_claim_score` table with the id and `display_order`
```
wtritch: nit: `display_order` => `data_order`

PR #313, opened 2018-12-26
add language table to reporting, olap

malaffoon:
Code changes to go with this: https://github.com/SmarterApp/RDW_Ingest/pull/554


reporting_olap/sql/V1_0_0_0__ddl.sql
```diff
@@ -38,6 +38,11 @@ CREATE TABLE staging_elas (
   code varchar(20) NOT NULL UNIQUE
```
malaffoon: For the previous schema change, i made it an incremental change to avoid having to remigrate all the data. However, for this change, there isn't enough data in redshift so we have to remigrate all the data. Since we have to do that, i reverted the previous delta and just baked it into this file.

warehouse/sql/V1_3_0_1__language.sql
```diff
@@ -546,8 +546,8 @@ UPDATE setting SET value = 'FALSE' WHERE name = 'AUDIT_TRIGGER_ENABLE' AND value
 
 -- do the work
 ALTER TABLE exam ADD COLUMN language_id smallint;
-CALL loop_by_exam_id('UPDATE exam e JOIN language l ON LOWER(e.language_code) = l.code SET e.language_id = l.id WHERE e.language_id IS NULL');
-CALL loop_by_exam_id('UPDATE exam e JOIN language l ON l.altcode IS NOT NULL AND e.language_code = l.altcode SET e.language_id = l.id WHERE e.language_id IS NULL');
+CALL loop_by_exam_id('UPDATE exam e JOIN language l ON LOWER(e.language_code) = l.code SET e.language_id = l.id, e.updated = e.updated WHERE e.language_code IS NOT NULL AND e.language_id IS NULL');
+CALL loop_by_exam_id('UPDATE exam e JOIN language l ON l.altcode IS NOT NULL AND e.language_code = l.altcode SET e.language_id = l.id, e.updated = e.updated WHERE e.language_code IS NOT NULL AND e.language_id IS NULL');
```
malaffoon: This is a tweak to the previous file to preserve the `updated` values.

PR #318, opened 2019-01-29
RP-84, adjust claim data order for ELA

malaffoon:
https://github.com/SmarterApp/RDW_Common/pull/90
https://github.com/SmarterApp/RDW_Ingest/pull/562

### DIFF
--- a/reporting/sql/V1_3_0_7__claim_order.sql
+++ b/reporting/sql/V1_3_0_7__claim_order.sql
@@ -1,0 +1,21 @@
+-- Fix the claim data order for ELA
+--
+-- The data order for ELA scorable claims has been Reading, Listening, Writing, Research.
+-- The middle two need to be flipped so it is Reading, Writing, Listening, Research.
+-- Because claims are denormalized in the reporting database we have to update asmt and exam.
+
+use ${schemaName};
+
+UPDATE subject_claim_score SET data_order = 2 WHERE subject_id = 2 AND code = '2-W';
+UPDATE subject_claim_score SET data_order = 3 WHERE subject_id = 2 AND code = 'SOCK_LS';
+
+UPDATE asmt SET claim2_score_code = '2-W', claim3_score_code = 'SOCK_LS'
+WHERE subject_id = 2 AND type_id IN (1,3) AND claim2_score_code = 'SOCK_LS' AND claim3_score_coode = '2-W';
+
+-- MySQL doesn't do swaps like other databases, so some trickery.
+-- This effectively swaps the three claim2/claim3 column values.
+UPDATE exam e JOIN asmt a on e.asmt_id = a.id SET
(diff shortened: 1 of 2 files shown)


reporting/sql/V1_3_0_7__claim_order.sql
```diff
@@ -0,0 +1,21 @@
...
+
+use ${schemaName};
+
+UPDATE subject_claim_score SET data_order = 2 WHERE subject_id = 2 AND code = '2-W';
+UPDATE subject_claim_score SET data_order = 3 WHERE subject_id = 2 AND code = 'SOCK_LS';
+
+UPDATE asmt SET claim2_score_code = '2-W', claim3_score_code = 'SOCK_LS'
+WHERE subject_id = 2 AND type_id IN (1,3) AND claim2_score_code = 'SOCK_LS' AND claim3_score_coode = '2-W';
+
+-- MySQL doesn't do swaps like other databases, so some trickery.
```
malaffoon: `UPDATE fubar SET a=b, b=a` does *not* do what you'd think. Or what PostgreSQL does. Or Oracle.

reporting/sql/V1_3_0_7__claim_order.sql
```diff
@@ -0,0 +1,21 @@
...
+-- This effectively swaps the three claim2/claim3 column values.
+UPDATE exam e JOIN asmt a on e.asmt_id = a.id SET
+  claim2_scale_score = @tmp_score := claim2_scale_score, claim2_scale_score = claim3_scale_score, claim3_scale_score = @tmp_score,
+  claim2_scale_score_std_err = @tmp_stderr := claim2_scale_score_std_err, claim2_scale_score_std_err = claim3_scale_score_std_err, claim3_scale_score_std_err = @tmp_stderr,
+  claim2_category = @tmp_cat := claim2_category, claim2_category = claim3_category, claim3_category = @tmp_cat
+WHERE a.subject_id = 2 AND a.type_id IN (1,3);
```
agorina: I am curious how long it will take on the large DB - do we need to batch it? Just a concern to consider.

agorina: And I am sure you thought about it - but just in case: we should back up the DB, or at least this table ( in case something goes wrong you we need to roll back).

malaffoon: In production there are only 415k records that will be affected. Since our batch size is typically 500k i didn't bother batching it. 
And, yeah, we always snapshot the databases before doing an upgrade. But i *really* hope that we don't have to roll anything back.

reporting/sql/V1_3_0_7__claim_order.sql
wtritch: That looks reasonable.  I seem to recall there being a place in the code where we manually controlled the display order as well...

PR #320, opened 2019-03-01
adds user query table

esotrope:


reporting/sql/V1_3_0_1__add_user_query_table.sql
```diff
@@ -0,0 +1,11 @@
+USE ${schemaName};
+
+CREATE TABLE IF NOT EXISTS user_query (
+  id bigint NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  user_login varchar(255) NOT NULL,
+  label varchar(255) NOT NULL,
+  query text NOT NULL,
```
esotrope: analogous to `user_report.report_request`

esotrope: didn't add `report_type/query_type` - do we use this in the `user_report` table for audit kind of info? i don't see it being used functionally for anything in the app

esotrope: thinking on my first pass i might drop label actually, wondering if it really isn't the same thing as the query name.

reporting/sql/V1_3_0_1__add_user_query_table.sql
```diff
@@ -0,0 +1,11 @@
+USE ${schemaName};
```
malaffoon: This is the first schema change as part of this phase, so i hadn't set it up yet. Can you bump the version in the build.gradle to `1.3.1-SNAPSHOT` and rename this file to `V1_3_1_1__add_user_query_table.sql`.

reporting/sql/V1_3_0_1__add_user_query_table.sql
malaffoon: Is the query name in the query text payload? If so, then i agree the extra label would be confusing and pointless. If not, then we need it somewhere. It would seem odd to me that the query text payload has the name, feels like it should be just the query. So we'd want the label as a column.

As for other columns ... having no other columns means that any presentation of the saved queries requires retrieving and parsing the query text. That may be okay because users won't have gazillions of saved queries but it feels a bit weird. But it might just be me overthinking it. Perhaps we can discuss.

esotrope: The query name is in the payload as it is part of the request to the server. I think it may make sense to have columns for fields that all queries share (which is probably only guaranteed to be the name). The reason for the json text column is to support polymorphism. if i recall right sql doesn't have conditional joins so it is a bit of a challenge to represent polymorphic entities without making them amorphic and introducing a catch-all table that supports all query fields right? but maybe that is actually what we want? not sure. we may want to pick Will's brain on this one a bit as he designed the original `user_report` table i think.

reporting/sql/V1_3_0_1__add_user_query_table.sql
malaffoon: @esotrope , i hope you see this as you make the changes ... i'm thinking this next phase should be 1.4.0 instead of 1.3.1. So let's make it `1.4.0-SNAPSHOT` and `V1_4_0_1__add_user_query_table.sql`. I'll change ingest as well, if you can change reporting. Thanks.

reporting/sql/V1_3_0_1__add_user_query_table.sql
malaffoon: Oh, i'm fine with the JSON payload as a text column. Just trying to think if we want any of the data broken out. I still think that saving the name separately makes sense, but we can discuss.

PR #322, opened 2019-03-23
add support for alt scores in the reporting schema

malaffoon:


reporting/sql/V1_4_0_2__alt_scoring.sql
```diff
@@ -0,0 +1,98 @@
...
+  code varchar(10) NOT NULL UNIQUE
+);
+
+INSERT INTO score_type (id, code) VALUES
+  (1, 'Overall'),
+  (2, 'Alt'),
+  (3, 'Claim');
+
+-- Note that we are keeping with the philosophy of the staging_ tables reflecting the warehouse
+-- tables, and doing the data transformations during the staging-to-reporting step.
```
malaffoon: This. It means that the `staging_` tables should look a lot like the warehouse tables (https://github.com/SmarterApp/RDW_Schema/blob/develop/warehouse/sql/V1_4_0_2__alt_scoring.sql).

reporting/sql/V1_4_0_2__alt_scoring.sql
```diff
@@ -0,0 +1,98 @@
...
+  ADD COLUMN alt2_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt3_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt4_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt5_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt6_score_code VARCHAR(10) NULL;
+
+-- this table isn't needed/used; its columns are denormalized into the asmt table
+DROP TABLE asmt_score;
+
+ALTER TABLE exam
```
malaffoon: I admit a bit of worry in the back of my mind: the exam table is wide. Most of the fields are small and some of them are most often null. But they are all pulled together so separating them into multiple 1-to-1 tables seems silly unless we encounter a database limitation. FTR, MySQL has a limit of 4096 columns and 64k bytes per row and we are now at about 70 columns and 1200 bytes.

reporting/sql/V1_4_0_2__alt_scoring.sql
```diff
@@ -0,0 +1,98 @@
...
+
+ALTER TABLE asmt
+  ADD COLUMN alt1_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt2_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt3_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt4_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt5_score_code VARCHAR(10) NULL,
+  ADD COLUMN alt6_score_code VARCHAR(10) NULL;
+
+-- this table isn't needed/used; its columns are denormalized into the asmt table
```
malaffoon: Finding this lying around surprised me a bit. @agorina any thoughts on why we kept it, perhaps just an oversight when we denormalized those columns?

agorina: Looks like an oversight. I will not be surprised if there are more cases like this.

reporting/sql/V1_4_0_2__alt_scoring.sql
agorina: I was just reading this: https://aws.amazon.com/blogs/database/planning-and-optimizing-amazon-aurora-with-mysql-compatibility-for-consolidated-workloads/

> All information in a given table row should be equally valuable. That is, you should query and manipulate data in all the fields of the row equally frequently. Storing hotter and colder data in the same row breeds inefficiencies.

malaffoon: Thanks Alla. I guess that lines up with this choice: currently we query all the columns in the table with equal frequency.

PR #324, opened 2019-03-29
add validation of student group count, by school and by school year

malaffoon:
Also a little cleanup.

### DIFF
--- a/validation/sql/student-groups-by-school/warehouse.sql
+++ b/validation/sql/student-groups-by-school/warehouse.sql
@@ -1,0 +1,5 @@
+SELECT d.name, d.natural_id, s.name, s.natural_id, count(*) AS count
+FROM student_group sg
+  JOIN school s on sg.school_id = s.id
+  JOIN district d on s.district_id = d.id
+GROUP BY sg.school_id;
--- a/validation/sql/student-groups-by-school/warehouse.sql
+++ b/validation/sql/student-groups-by-school/warehouse.sql
@@ -2,4 +2,5 @@ SELECT d.name, d.natural_id, s.name, s.natural_id, count(*) AS count
 FROM student_group sg
   JOIN school s on sg.school_id = s.id
   JOIN district d on s.district_id = d.id
+WHERE sg.deleted = 0
 GROUP BY sg.school_id;
--- a/validation/sql/total-ica-scores/reporting.sql
+++ b/validation/sql/total-ica-scores/reporting.sql
@@ -1,6 +1,6 @@
 SELECT
     sum(scale_score),
-    sum(scale_score_std_err),
+    round(sum(scale_score_std_err), 1),
     sum(performance_level)
   FROM exam
   WHERE type_id = 1 AND scale_score IS NOT NULL;
--- a/validation/validate-migration.sh
+++ b/validation/validate-migration.sh
@@ -33,12 +33,12 @@ function print_connection() {
 
 function mysql_to_csv() {
     declare -a connection=("${!1}")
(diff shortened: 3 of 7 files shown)


validation/sql/student-groups-by-school/warehouse.sql
```diff
@@ -0,0 +1,5 @@
+SELECT d.name, d.natural_id, s.name, s.natural_id, count(*) AS count
+FROM student_group sg
+  JOIN school s on sg.school_id = s.id
+  JOIN district d on s.district_id = d.id
```
malaffoon: oops, need a WHERE deleted=0 here

validation/sql/total-ica-scores/reporting.sql
```diff
@@ -1,6 +1,6 @@
 SELECT
     sum(scale_score),
-    sum(scale_score_std_err),
+    round(sum(scale_score_std_err), 1),
```
malaffoon: unrelated; eliminate excess precision because it was *always* indicating a diff, usually at the 3rd or 4th decimal.

validation/validate-migration.sh
```diff
@@ -56,8 +56,8 @@ function run_test() {
     local test_headers=`get_property_by_index $3 2`
     declare -a connection=("${!4}")
     local csv_file=`mktemp`
-    echo "${test_headers}" >>  ${csv_file}
-    if [ "${database_type}" == "mysql" ]; then
+    echo "${test_headers}" >> ${csv_file}
+    if [[ "${database_type}" == "mysql" ]]; then
```
malaffoon: IJ (well, the bash plugin) warns with single brackets. I didn't research it, i just let IJ fix all.

validation/validate-migration.sh
```diff
@@ -148,41 +146,41 @@ function print_summary() {
...
+[[ ${run_olap} == 1 ]] && [[ -z "${reporting_olap_user}" ]] && echo "reporting_olap_user must be set" && exit 1
+[[ ${run_olap} == 1 ]] && [[ -z "${reporting_olap_password}" ]] && echo "reporting_olap_password must be set" && exit 1
 
 # settings
 start_time=`now_in_seconds`
 base_dir=`cd "$(dirname "$0")" ; pwd -P`
 sql_dir=${base_dir}/sql
 out_dir="${base_dir}/results-$(now_in_YYYY_mm_dd_HHMMSS)"
-diff_options="-y --suppress-common-lines"
+diff_options="-y --suppress-common-lines -W 200"
```
malaffoon: The new validations are a bit longer i guess ... this avoids truncating the useful count information in the diff.

malaffoon:
I'm going to merge so i can test in production. I will still address any comments, of course.

validation/sql/student-groups-by-school/warehouse.sql
```diff
@@ -0,0 +1,6 @@
...
+WHERE sg.deleted = 0
+GROUP BY sg.school_id;
```
agorina: Do we need to include `active=0` too?

malaffoon: Well shoot, i'm reviewing the migrate script and it seems inconsistent. For the student_group_membership and user_student_group it checks `wsg.deleted = 0 AND wsg.active != 0` but, for the student_group itself it checks neither. This seems wrong to me. @agorina any reason the master table would not have the check while the child table does? Were we being clever knowing that a group with no users or students is effectively hidden? Or is it just an oversight (because such groups are effectively hidden)?

agorina: My best guess (without looking at the code) is that the master table would need both active/inactive to know when the `active` flag changes.

PR #327, opened 2019-04-13
make translation labels unlimited length

malaffoon:


reporting/sql/V1_4_0_3__translation_label.sql
```diff
@@ -0,0 +1,6 @@
+-- Modify translation label to be TEXT, i.e. no length limit
+
+use ${schemaName};
+
+ALTER TABLE staging_accommodation_translation
+  MODIFY COLUMN label TEXT NOT NULL;
```
malaffoon: It was VARCHAR(40). Technically that matches the SB spec but having any limit on a translation is dangerous. And, sure enough, 2019 translations are up to 65 characters.

PR #329, opened 2019-05-14
adds input type column to pipeline table and default empty pipeline s…

esotrope:
…cripts

### DIFF
--- a/warehouse/sql/V1_4_0_5__pipeline_input_type.sql
+++ b/warehouse/sql/V1_4_0_5__pipeline_input_type.sql
@@ -1,13 +1,9 @@
--- Adds input type column to pipeline table
--- Sets pipeline input types
--- Adds default empty script for each pipeline
+-- Adds input type column to pipeline table and sets each pipelines input types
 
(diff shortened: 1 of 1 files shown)


warehouse/sql/V1_4_0_5__pipeline_input_type.sql
```diff
@@ -0,0 +1,25 @@
...
+UPDATE pipeline
+  SET input_type = 'csv'
+WHERE code in ('group', 'assessment');
+
+-- provide pipelines with an empty script
+
+INSERT INTO pipeline_script (pipelineId, body, updated_by) VALUES
+ ((SELECT id FROM pipeline WHERE code = 'exam'), '', 'System'),
+ ((SELECT id FROM pipeline WHERE code = 'group'), '', 'System'),
+ ((SELECT id FROM pipeline WHERE code = 'assessment'), '', 'System');
```
esotrope: should we just blank this out and not show it in the UI or actually show that it was the system default? thoughts?

PR #330, opened 2019-05-15
adds pipeline active version column

esotrope:


warehouse/sql/V1_4_0_6__pipeline_active_version.sql
```diff
@@ -0,0 +1,10 @@
+-- Adds active version column to pipeline table
+-- Makes pipeline input_type column not null
+
+use ${schemaName};
+
+ALTER TABLE pipeline
+  ADD COLUMN active_version int;
+
+ALTER TABLE pipeline
+  MODIFY input_type varchar(20) NOT NULL;
```
esotrope: forgot to make this one non null - or can it be null?

malaffoon: This seems right, the expected input type should be known for all pipelines. If a pipeline can have multiple input types, we might need to do something but this is good for now.

PR #331, opened 2019-05-26
Feature/jar tooling

ftdholbrook:
I couldn't find a good way to include the migrations as they were in the top level directory, so I added a sub-directory called "migrations".  I updated the gradle tasks to match.


src/main/java/org/opentestsystem/rdw/schema/SchemaMigration.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Wraps {@see org.flywaydb.core.Flyway#repair()}
+     *
+     * @throws FlywayException when the schema history table repair failed.
+     */
+    public void repair() throws FlywayException {
+        flyway.repair();
+    }
```
ftdholbrook: I don't have tests for, or a use case for basline() and repair() I included them because they were in the gradle scripts.  I'm open to removing them.

src/test/java/org/opentestsystem/rdw/schema/SchemaMigrationWarehouseTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.opentestsystem.rdw.schema.SchemaMigration.Migration.Warehouse;
+
+public class SchemaMigrationWarehouseTest {
```
ftdholbrook: This is only testing Warehouse / Mysql I think it is enough to prove that it is working.

build.gradle
```diff
@@ -116,16 +130,24 @@ flywayActions.keySet().each { action ->
...
-                include 'reporting/**'
-                include 'warehouse/**'
-                include 'migrate_olap/**'
-                include 'reporting_olap/**'
-                include 'legacy_load/**'
+                include 'migrations/reporting/**'
+                include 'migrations/warehouse/**'
+                include 'migrations/migrate_olap/**'
+                include 'migrations/reporting_olap/**'
+                include 'migrations/legacy_load/**'
```
malaffoon: really don't need the `legacy_load` stuff

src/main/java/org/opentestsystem/rdw/schema/SchemaMigration.java
malaffoon: They are nice to have in the zip artifact, don't see any reason to have them available for java invocation.

PR #332, opened 2019-06-04
consolidate sql scripts

malaffoon:
It's that time of year ...


migrations/warehouse/sql/V1_4_0_0__update.sql
```diff
@@ -0,0 +1,183 @@
...
+
+
+ALTER TABLE accommodation_translation
+    MODIFY COLUMN label TEXT NOT NULL;
+
+CREATE TABLE IF NOT EXISTS pipeline (
+    id tinyint NOT NULL PRIMARY KEY,
+    code varchar(20) NOT NULL UNIQUE,
+    input_type varchar(20) NOT NULL,
+    active_version int
```
esotrope: in practice this is an int but the java modeling is lenient enough to use string, should we change this to varchar? or just leave it?

malaffoon: Good point, i'll change to a varchar

PR #333, opened 2019-06-05
Add PIPELINE_FAILURE to import_status

gregcharles:
Is this right, or should it also be consolidated to the 1.4.0 script?


malaffoon:
Gah, this always happens ... right after i consolidate the scripts a new thing needs to be added. This is right, but i will try to consolidate it right away.

PR #335, opened 2019-06-14
workaround for redshift specific migrations

ftdholbrook:


ftdholbrook:
This workaround will not work

PR #337, opened 2020-05-27
Update README

vgonzalez9:
Updating the location of the my.cnf file that is modified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ echo 'export PATH="/usr/local/opt/mysql@5.6/bin:$PATH"' >> ~/.zshrc
 brew services start mysql@5.6
 mysql_tzinfo_to_sql /usr/share/zoneinfo | mysql -u root mysql
 ```
-Edit `/usr/local/Cellar/mysql@5.6/5.6.*/my.cnf` and add the following lines:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,7 +37,7 @@ echo 'export PATH="/usr/local/opt/mysql@5.6/bin:$PATH"' >> ~/.zshrc
 brew services start mysql@5.6
 mysql_tzinfo_to_sql /usr/share/zoneinfo | mysql -u root mysql
 ```
-Edit `/usr/local/Cellar/mysql@5.6/5.6.*/my.cnf` and add the following lines:
+Edit `/usr/local/etc/my.cnf` and add the following lines:
```
malaffoon, line 40: This isn't wrong but i'm curious if you found advice to modify that my.cnf over the one in the cellar? Because the location of the file is different depending on your installation, it might be worth documenting the fs_usage trick:

To determine the location of the `my.cnf` file, you can monitor for the usage of the file while restarting mysql. Open two terminal windows. In one do `sudo fs_usage | grep my.cnf`. In the second terminal, restart mysql: `brew services restart mysql@5.6`. In the first window you'll see a number of locations being accessed, and one or more of these will actually exist. You can edit any of those.

PR #343, opened 2020-06-15
insert trait purpose/category labels for ELA

malaffoon:
Greg, i realized that we need a subject-specific place to specify the labels for the purposes and categories. I know that, in theory, the way i've done the categories isn't as flexible as it could be but i'm leaving it this way for now, let me know if you think we should change it. Specifically, i have `subject.ELA.trait.category.ORG` where i should probably have the purpose in there, something like `subject.ELA.trait.purpose.ARGU.category.ORG`. The reason i left it as is, is that i am imagining a fail-over approach: first look up the most specific label (`...trait.purpose.ARGU.category.ORG`) then fallback to a less specific label (`...trait.category.ORG`)


src/test/java/org/opentestsystem/rdw/schema/SchemaMigrationWarehouseTest.java
```diff
@@ -17,17 +17,17 @@
```
malaffoon, line 17: Unrelated warning cleanup.

PR #344, opened 2020-07-08
add lifecycle message for flyway tasks; add infoAll tasks

malaffoon:
It always freaks me out a little when i'm applying the flyway tasks to other datasources. This outputs the database url and schema as part of the task so you can do something like:
```
gw infoWarehouse
> Task :infoWarehouse
Running infoWarehouse on jdbc:mysql://localhost:3306/warehouse
Schema version: 1.4.0.0
...
```


gregcharles:
Sorry, late the party, but that looks good. I'll try it out.

PR #346, opened 2020-07-23
ISR support for exam level trait scores

malaffoon:
add WER mapping table; add text entries for ISRs; add script for converrting summative item trait scores to exam level trait scores

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -109,7 +109,7 @@ flywayActions.keySet().each { action ->
                 group = 'Schema'
                 description = description + ' (url: ' + config[2] + ', schema: ' + config[3] + ')'
                 doFirst {
-                    logger.lifecycle(MessageFormat.format("Running {0} on {1}{2}", taskName, config[2], config[3]))
+                    logger.lifecycle(MessageFormat.format("Running {0} on {1}, {2}", taskName, config[2], config[3]))
                 }
             }
         }
--- a/migrations/reporting/sql/V2_4_0_3__wer_purpose.sql
+++ b/migrations/reporting/sql/V2_4_0_3__wer_purpose.sql
@@ -1,0 +1,34 @@
+-- v2.4.0.3 flyway script
+--
+-- add lookup table for WER types
+-- additional subject translation entries
+
+use ${schemaName};
+
+-- reference table for hard-coded mapping of WER item type to purpose
+CREATE TABLE wer_purpose (
+  wer_type varchar(20) not null,
+  purpose varchar(10) not null
+);
+
+INSERT INTO wer_purpose (wer_type, purpose) VALUES
+  ('Argumentative', 'ARGU'),
+  ('Explanatory', 'EXPL'),
+  ('Informative', 'INFO'),
+  ('Narrative', 'NARR'),
+  ('Opinion', 'OPIN');
+
+INSERT INTO subject_translation (subject_id, label_code, label) VALUES
+  (2, 'subject.ELA.trait.category.CON.0', 'The response demonstrates little or no command of conventions.'),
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -109,7 +109,7 @@ flywayActions.keySet().each { action ->
                 group = 'Schema'
                 description = description + ' (url: ' + config[2] + ', schema: ' + config[3] + ')'
                 doFirst {
-                    logger.lifecycle(MessageFormat.format("Running {0} on {1}{2}", taskName, config[2], config[3]))
+                    logger.lifecycle(MessageFormat.format("Running {0} on {1}, {2}", taskName, config[2], config[3]))
```
malaffoon, line 112: unrelated, just makes it easier to read

migrations/reporting/sql/V2_4_0_3__wer_purpose.sql
```diff
@@ -0,0 +1,34 @@
...
+);
+
+INSERT INTO wer_purpose (wer_type, purpose) VALUES
+  ('Argumentative', 'ARGU'),
+  ('Explanatory', 'EXPL'),
+  ('Informative', 'INFO'),
+  ('Narrative', 'NARR'),
+  ('Opinion', 'OPIN');
+
+INSERT INTO subject_translation (subject_id, label_code, label) VALUES
```
malaffoon, line 21: Turns out the ISR has text specific to the category score. I've added these as additional messages in the subject definition file; not sure how to get that into the subject workbook, but we might need to.

migrations/reporting/sql/V2_4_0_3__wer_purpose.sql
```diff
@@ -0,0 +1,34 @@
...
+
+use ${schemaName};
+
+-- reference table for hard-coded mapping of WER item type to purpose
+CREATE TABLE wer_purpose (
+  wer_type varchar(20) not null,
+  purpose varchar(10) not null
+);
+
+INSERT INTO wer_purpose (wer_type, purpose) VALUES
```
malaffoon, line 14: This table serves as a reference table for posterity, and it also makes the rollup query simpler. That will be part of a RDW_Reporting checkin.

migrations/reporting/sql/V2_4_0_3__wer_purpose.sql
gregcharles: Are they supposed to go in the UI? How ... by hovering over each score?

migrations/reporting/sql/adjust_summative_trait_scores.sql
```diff
@@ -0,0 +1,91 @@
+-- Before 2.4, writing trait scores were associated with items. For the generated data this worked
+-- for both summative and interim tests. However, in the wild, summative tests don't have any item
+-- data at all. To display writing trait scores, the UI would (?verify?) iterate through the items
+-- and aggregate the scores, presenting them in a table. For the ISR's, the item level data was
+-- rolled up and stored at the exam level. The trait category was hard-coded and denormalized: each
+-- item has three columns, one for each category. The trait purpose was stored with the assessment
+-- item definition.
+--
```
gregcharles, line 8: (re: ?verify?) Prior to this phase anyway, the UI worked with the same data structure for summatives as it did for interims. The only difference in handling was that the summatives didn't get a totals row. Any other differences must have been happening on the back end where that data structure was created. However, as we discovered, only the test data sets regularly had this data for summative reports. For most cases in the UAT data set, there were no writing traits available for ELA summative reports.

migrations/reporting/sql/V2_4_0_3__wer_purpose.sql
malaffoon: These are in the ISR (printed report).

PR #348, opened 2020-08-18
RDW6, req1: Updates for subject in embargo audit log.

gregcharles:


migrations/warehouse/sql/V2_4_0_5__embargo_audit.sql
```diff
@@ -0,0 +1,67 @@
+-- v2.4.0_4 flyway script
```
malaffoon, line 1: V2.4.0_5?

migrations/warehouse/sql/V2_4_0_5__embargo_audit.sql
```diff
@@ -0,0 +1,67 @@
...
+    ADD CONSTRAINT fk__audit_district_embargo__subject FOREIGN KEY (subject_id) REFERENCES subject(id),
+    ADD INDEX idx__audit_district_embargo__subject (subject_id),
+    ADD CONSTRAINT fk__audit_district_embargo__district FOREIGN KEY (district_id) REFERENCES district(id),
+    ADD INDEX idx__audit_district_embargo__district (district_id);
+
+DROP TRIGGER trg__district_embargo__insert;
+DROP TRIGGER trg__district_embargo__update;
+DROP TRIGGER trg__district_embargo__delete;
+
+CREATE TRIGGER trg__district_embargo__insert
```
malaffoon, line 21: I don't see this joining in the audit setting. Did we discuss whether this should honor that setting like the other audit tables?

migrations/warehouse/sql/V2_4_0_5__embargo_audit.sql
gregcharles: 👍

migrations/warehouse/sql/V2_4_0_5__embargo_audit.sql
gregcharles: Do you mean the addition of the foreign keys? I need the subject and district names in the audit log, but should I do it a different way?

malaffoon: If you look in the other triggers, you'll see they join the settings table and only insert the audit records if the audit flag is enabled.

gregcharles: Ah, got it. I didn't know that. The embargo audit triggers weren't like that before, but I can make them like that now. Should I? I don't know why they would be different, but maybe they just predate the settings thing and weren't retrofitted.

malaffoon: I could see an argument either way. But, if they weren't like that before, let's not complicate it.

PR #349, opened 2020-08-19
better handle existing entries in the district_embargo table

malaffoon:


migrations/warehouse/sql/V2_4_0_4__test_results_availability.sql
```diff
@@ -1,18 +1,35 @@
 -- v2.4.0_4 flyway script
 --
--- adds support for finer-grained embargo control
+-- NOTE: this script was modified after initial creation which will mess up
+-- things if the older version was run successfully. The original version had
+-- a checksum of -385633165. However, if it was run successfully, the embargo
+-- table must've been empty which means you don't need to rerun this script.
+-- Update the schema_version and set the checksum to the correct value for
+-- this version of the script (which can't be shown here because documenting
```
malaffoon, line 8: I will document the two checksum values in the upgrade notes

gregcharles: Hmm, so it wasn't what I thought, because the subject_id was nullable. You added migration of old records, which I hadn't done yet, but why did it error? Is one of the indexes unhappy with null values?

malaffoon: Foreign key on the subject_id field: there is no null id entry in the subject table. I believe there is a way to tell it to ignore the foreign key for null values, but that isn't really what we want.

gregcharles: Hmm, interesting. Maybe it is what we want for audit_district_embargo though. I added a foreign key to it as well in order to be able to retrieve subject codes with the audit record. I don't think we'll want to migrate the old audit records though. IOW, I think there's going to be an issue with 2_4_0_5 as well. :(

malaffoon: Hmm. The district_embargo table has subject_id defined as default value = 0. That seems weird because it isn't in the script anywhere. The audit_district_embargo table has subject_id with no default value, and null entries work just fine.  Gah.

I do agree that we shouldn't bother with migrating the old audit table entries.

PR #350, opened 2020-08-24
RDW5, req1 -- exam counts table for embargo

gregcharles:
Adds exam count table, but doesn't populate it yet. The unpopulated counts will appear as 0 in the test results availability (embargo) results.

### DIFF
--- a/migrations/warehouse/sql/V2_4_0_6__embargo_exam_counts.sql
+++ b/migrations/warehouse/sql/V2_4_0_6__embargo_exam_counts.sql
@@ -1,0 +1,16 @@
+-- v2.4.0_6 flyway script
+--
+-- adds exam_count table, which will hold pre-calculated summative exam counts by school year, district, and subject
+-- for display in the test results availability (embargo) UI.
+--
+use ${schemaName};
+
+CREATE TABLE exam_count
+(
+    school_year SMALLINT NOT NULL,
+    district_id INT NOT NULL,
+    subject_id SMALLINT NOT NULL,
+    count INT DEFAULT 0 NOT NULL,
+    PRIMARY KEY (school_year, district_id, subject_id)
+);
+


migrations/warehouse/sql/V2_4_0_6__embargo_exam_counts.sql
```diff
@@ -0,0 +1,16 @@
...
+
+CREATE TABLE exam_count
+(
+    school_year SMALLINT NOT NULL,
+    district_id INT NOT NULL,
+    subject_id SMALLINT NOT NULL,
+    count INT DEFAULT 0 NOT NULL,
+    PRIMARY KEY (school_year, district_id, subject_id)
+);
+
```
gregcharles, line 16: I don't think foreign keys are needed here since this table will not be joined directly with district or subject. My benchmarking of the new embargo query (PR to follow in RDW_Reporting) seemed to bear this out.

malaffoon: The foreign keys aren't, in general, there for performance (that's what the auto-created associated indexes are for) ... they are there to protect against invalid entries. I would be happier if there were foreign keys to enforce integrity. I won't lose sleep over it though.

gregcharles: Hmm, regular indexes perform as well as foreign keys for joining? Someday I'll have to test that. I don't really get why we have the associated indexes for all the foreign keys. I hadn't even noticed that before a couple of weeks ago. Does having both indexes give an advantage?

malaffoon: Read Alla's comment at the top of V1_0_0_0__ddl.sql, it explains why we're explicitly creating the indexes. Foreign keys, by themselves, do absolutely nothing for performance of queries.

PR #351, opened 2020-09-04
schema changes to support phase 6 embargo changes in reporting

malaffoon:
I think the schema changes are complete but won't know for sure until all the coding is complete and tested. These changes should *not* be merged until just before the code changes in RDW_Reporting and RDW_Ingest are ready to commit.


migrations/reporting/sql/V2_4_0_4__test_results_availability.sql
```diff
@@ -0,0 +1,61 @@
+-- v2.4.0_4 flyway script
+--
+-- adds support for finer-grained embargo control
+
+use ${schemaName};
+
+-- reference table
```
malaffoon, line 7: Technically not necessary, but this makes it consistent with other reference values (e.g. asmt_type). Convenient to have these values stored in the database.

migrations/reporting/sql/V2_4_0_4__test_results_availability.sql
```diff
@@ -0,0 +1,61 @@
...
+  DROP COLUMN embargo_enabled;
+
+-- recreate the staging table
+DROP TABLE staging_district_embargo;
+CREATE TABLE staging_district_embargo (
+  district_id int NOT NULL,
+  school_year smallint NOT NULL,
+  subject_id smallint NOT NULL,
+  individual tinyint,
+  migrate_id bigint NOT NULL,
```
malaffoon, line 59: Previously, the staging table had both the `individual` and `aggregate` field. I don't see why the `aggregate` field is needed in the reporting database, so i removed it. Once code changes are complete it can be confirmed as to whether it should be here or not.

migrations/reporting_olap/sql/V2_4_0_4__test_results_availability.sql
```diff
@@ -0,0 +1,61 @@
...
+
+-- recreate staging table
+DROP TABLE staging_district_embargo;
+CREATE TABLE staging_district_embargo (
+  district_id integer NOT NULL,
+  school_year smallint NOT NULL,
+  subject_id smallint NOT NULL,
+  aggregate smallint,
+  UNIQUE (district_id, school_year, subject_id)
+);
```
malaffoon, line 61: I'm pretty sure there will be another change to this script: there is a state_embargo table that may need to be removed. I haven't looked at the OLAP code to know if it will still be needed, so i didn't make that change yet.

migrations/reporting/sql/V2_4_0_4__test_results_availability.sql
gregcharles: 👍

migrations/reporting/sql/V2_4_0_4__test_results_availability.sql
gregcharles: That surprised me. What about aggregate reports? Isn't their visibility based on aggregate embargo status? I feel like the answer is, well, duh, those look at the warehouse schema. Is that right?

migrations/reporting_olap/sql/V2_4_0_4__test_results_availability.sql
gregcharles: Thanks! I made a note of it.

gregcharles (approved):
lgtm

migrations/reporting/sql/V2_4_0_4__test_results_availability.sql
malaffoon: All aggregate reports are generated from the reporting_olap schema which does, as you note, get the `aggregate` flag migrated to it.

PR #370, opened 2022-02-23
Update student upsert

norway77:
please look this over.  I need to add a before field as well before merge


xuping-xu:
Let us change this sql script to V2_5_0_10. And we can make any changes related to Lexile/Quantile in this branch. But currently let us not merge the changes to the develop branch.  

In addition to adding the lexile_quantile column to the warehouse.exam table, we also need to add it to the following tables: warehouse.audit_exam, warehouse.audit_exam_score, warehouse.exam_score, and reporting.exam, reporting.staging_exam, reporting.staging_exam_score
We can do this in the same V2_5_0_10 sql script. Also, it would be nice if we can order this lexile_quantile column to be right after the performance_level column (so as to be consistent with the TRT and make the column easier to locate).

At this time, let us create another branch from the develop branch and start working on V2_5_0_9 for adding the latest_school_update column.

norway77:
ok, looks good

PR #371, opened 2022-03-02
Feature/srs 936

xuping-xu:
I have cleaned up the stored procedure and added comments so as to improve its readability. I have performed the warehouse migration on my local database and run the stored procedure and it seems working. 

You may test it on your local machine and let me know your feedback.


migrations/warehouse/sql/V2_5_0_10__student_upsert_changes.sql
```diff
@@ -0,0 +1,111 @@
...
+    DECLARE CONTINUE HANDLER FOR 1062
+    BEGIN
+        SELECT id, 0 INTO p_id, p_updated FROM student WHERE ssid = p_ssid;
+    END;
+
+    -- In the unlikely case where p_exam_completed_at is still null (unlikely because the pipeline script
+    -- would set dateCompleted by statusDate value if dateCompleted is blank), default the exam completion
+    -- time to current timestamp.
+    SELECT CASE WHEN p_exam_completed_at IS NULL THEN CURRENT_TIMESTAMP(6) ELSE p_exam_completed_at END
+    INTO p_exam_completed_at;
```
malaffoon, line 37: In general it is bad style to modify IN variables, and since it is, in your own words, unlikely, perhaps use an IF block instead?
IF (p_exam_completed_at IS NULL) THEN
   SELECT CURRENT_TIMESTAMP(6) INTO p_exam_completed_at;
END IF;

xuping-xu: @malaffoon Totally agree. I greatly appreciate your comments.

BTW: For some reason, although I intend to request review from @norway77 , the request also comes to you. Sorry about that. I will double check to make sure that this would not be the case next time. However, I really appreciate your insightful comments. Thank you very much!

norway77: thank you Mark.  Xuping looks good

